PR #31, opened 2019-08-01
Update host address in jscoresdk

BBTN:


autotest/client/jscoresdk/utils/constants.ts
```diff
@@ -47,7 +47,7 @@ export const config = {
     productID: 'JSCoreSDKTest',
     version: '1.0.0',
     component: 'sunglow',
-    host: 'https://localhost:8090/',
+    host: 'http://localhost:8091/',
     language: 'en-US',
```
nannany00: It uses the API that getPatternbyLocale, not by language and region, right?

BBTN: As we discussed, this commit doesn't have influence on the API.

PR #35, opened 2019-08-01
 remove '[FATAL_ERROR]' from log 

linr211:


g11n-ws/modules/md-service-i18n-l3/src/main/java/com/vmware/vip/core/messages/service/product/ProductService.java
```diff
@@ -90,8 +88,7 @@ public DropVersionDTO getVersionInfo(String productName, String version)
 		try {
 			jsonStr = productdao.getVersionInfo(productName, version);
 		} catch (DataException e) {
-			throw new L3APIException(ConstantsKeys.FATA_ERROR
-					+ "Failed to drop version info for " + productName
+			throw new L3APIException("Failed to drop version info for " + productName
```
nannany00: It should be "Failed to get drop version info for", right?

g11n-ws/modules/md-service-i18n-l3/src/main/java/com/vmware/vip/core/messages/service/product/ProductService.java
```diff
@@ -61,8 +61,7 @@
 				supportedLocaleList.remove(ConstantsKeys.LATEST);
 			}
 		} catch (DataException e) {
-			throw new L3APIException(ConstantsKeys.FATA_ERROR
```
surite: why remove FATAL error?

linr211: Because it's not a fatal error as it can't detect if the bundle is lost or not.

linr211: Especially for new product.

PR #101, opened 2019-08-13
Add the static pages localization sample. 

nikichak:
Add the static pages localization solution in the sample folder. 
Use the core SDK client. 
Add util that provides functionality for translating a static page. 
Add source bundle and translation_zh.json in order to not call the backend.
Add the attribute l10n and the key of the message that has to be translated as its value. 
Load the language either from the cookies or from the browser, if not supported fall back to english.


samples/client/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+$ npm start
+```
+
+open: http://localhost:3000/example
+
+# In the dev tools terminal change the lang
+
+```
+$ document.cookie = "local=zh"
```
pengrongbo: Type error

pengrongbo: local=zh

samples/client/README.md
```diff
@@ -0,0 +1,40 @@
+# Prerequisites
+
+- Build and run the [Singleton Service](https://github.com/vmware/singleton).
+
+> Note:This sample application has been preconfigured to work with default Singleton service settings. Change the configurations in [i18n.utils.js](https://github.com/vmware/singleton/tree/g11n-js-client/samples/client/src/i18n.utils.js) as needed.
+
+> Note:Language and Region can be changed by adding the locale in the cookie with name 'vcs_locale' (https://github.com/vmware/singleton/tree/g11n-js-client/samples/client/src/i18n.utils.js). Currently, only Chinese and English are supported in this sample application.
```
pengrongbo: the cookie name doesn't match the code

samples/client/src/i18n.utils.js
```diff
@@ -0,0 +1,78 @@
...
+
+    let i18nClient = vipCore.i18nClient.init(
+        {
+            productID: 'CoreSDKClient',
+            version: '1.0.0',
+            component: 'ui',
+            host: 'http://localhost:8091',
+            language: lang,
+            sourceBundle: ENGLISH,
+            i18nAssets: 'src/sources/',
```
pengrongbo: core SDK doesn't support 'i18nAssets, we may need to remove it.

samples/client/src/sources/translation_zh.json
```diff
@@ -0,0 +1,27 @@
+{
```
pengrongbo: what's this for?

samples/client/js/test/i18n.utils.test.js
```diff
@@ -0,0 +1,83 @@
+import { localize } from "../src/i18n.utils";
+import * as vipCore from '@vip/vip-core-sdk';
+import { config as defaultConfig } from "../src/config";
+import { ENGLISH } from "../src/source.l10n";
+
+describe("Localize static files solution tests", () => {
+
+    describe("Pseudo extraction tests", () => {
+
+        it("should call the localStoraeg with the default config key", () => {
```
wangyi-e: localStoraeg, spelling

samples/client/src/i18n.utils.js
nchakarovarkin: It actually supports

samples/client/js/test/i18n.utils.test.js
nchakarovarkin: Fixed

PR #113, opened 2019-08-16
Using the function parameters instead of current language or region, …

pengrongbo:
- shouldn't use the current language or region to query cache


junw-33455432:
To merge the PR, all commits in PR need to be signed, please refer to 'Set Signature' in https://github.com/vmware/singleton/blob/master/CONTRIBUTING.md

PR #117, opened 2019-08-16
Enable withCredentials

pengrongbo:
- Add HTTP options object to pass the parameters
- For now, just support customizing timeout and credentials

### DIFF
--- a/src/services/core.service.ts
+++ b/src/services/core.service.ts
@@ -2,7 +2,7 @@
  * Copyright 2019 VMware, Inc.
  * SPDX-License-Identifier: EPL-2.0
  */
-import { Loader } from '../loader';
+import { Loader, HttpRequestOptions } from '../loader';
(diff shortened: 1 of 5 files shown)


src/services/core.service.ts
```diff
@@ -32,6 +33,7 @@ export class CoreService {
         this.config.sourceBundle = this.resetSourceBundle();
         this.currentLanguage = vipConfig.language;
         this.currentRegion = vipConfig.region;
+        this.httpOptions = vipConfig.httpOptions;
```
wangyi-e: L30: only the ‘withCredentials’ passed in httpOptions,  timeout will be overridden.

pengrongbo: Good concern! The default timeout is set to 0 which indicates no timeout. With that default value, any remote end can keep us waiting for the requested resource for an indefinite period.

PR #118, opened 2019-08-19
Fix timeout issue

wangyi-e:
- If timeout not defined in init config, should filled with default value.
- Add code comment.
- Update readme.

### DIFF
--- a/src/loader.ts
+++ b/src/loader.ts
@@ -101,7 +101,12 @@ export class VIPLoader implements I18nLoader {
         const url = this.getSupportedRegionsUrl(language, conf.host);
         return this.getRequest(url, conf.timeout, (res: Object) => this.responseParser.ParseSupportedRegionsData(res));
     }
-
+    /**
+     * Get request.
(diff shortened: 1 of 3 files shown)


src/loader.ts
```diff
@@ -115,6 +120,13 @@ export class VIPLoader implements I18nLoader {
             );
     }
 
+    /**
+     * Post request.
```
pengrongbo: Fetch I18N resource from backend service through post request.

src/loader.ts
```diff
@@ -101,7 +101,12 @@ export class VIPLoader implements I18nLoader {
         const url = this.getSupportedRegionsUrl(language, conf.host);
         return this.getRequest(url, conf.timeout, (res: Object) => this.responseParser.ParseSupportedRegionsData(res));
     }
-
+    /**
+     * Get request.
```
pengrongbo: Fetch I18N resource from backend service through get request.

PR #126, opened 2019-08-21
Update issue templates

gong-yu:


nannany00:
Will the template be chosen, I see there is a drop down list to chose a template when create an issue? or it's only a standard to create issue?

PR #128, opened 2019-08-22
Upgrade jackson-databind from 2.9.9 to 2.10.0.pr1 and Upgrade swagger-ui

tigershi:
Upgrade jackson-databind from 2.9.9 to 2.10.0.pr1 and Upgrade swagger-ui


tigershi:
The issues have combined to one issue

PR #143, opened 2019-08-27
Support adding extra headers to HTTP request

Xiaochao8:


src/main/java/com/vmware/vipclient/i18n/base/VIPService.java
```diff
@@ -77,4 +78,8 @@ public String getProductID() {
 	public String getVersion() {
 		return version;
 	}
+	
+	public void setHeaderParams(Map<String, String> params) {
```
linr211: it's better to check if httpRequester is null before set the values.

src/main/java/com/vmware/vipclient/i18n/base/HttpRequester.java
```diff
@@ -269,8 +278,17 @@ public void checkServerTrusted(X509Certificate[] cert,
...
 
 	public boolean isConnected() {
-		String ipAddress = (vipHostName.contains(":")) ? vipHostName.split(":")[0]
-				: vipHostName;
+		String ipAddress = (vipHostName.contains(":")) ? vipHostName.split(":")[0] : vipHostName;
 		return HttpRequester.ping(ipAddress);
 	}
+
+	private void addHeaderParams(HttpURLConnection connection) {
+		if (null == customizedHeaderParams) {
```
linr211: it's also better to check if connection is null. so please change it to this?
if (null == customizedHeaderParams || connection == null)

Issue #149, opened 2019-08-29
[Singleton Service][Get pattern by locale] return default region when input is a language


nannany00:
Singleton Service API:
GET /i18n/api/v2/formatting/patterns/locales/{locale} (Get pattern from CLDR)

This API is used to provide the pattern data for dates/numbers/currencies/plurals/measurements from CLDR according to the requested locale.

------
The issue is:
When the request locale is a language without default region, please add the default region in Response.

Example:
locale=en; scope=numbers

Current Result:
The region is "" for en.
```
...
 "data": {
    "localeID": "en",
    "language": "en",
    "region": "",
    "categories": {
      "numbers": {
        "numberSymbols": {
....
```

Suggestion Result:
Add default Region **US** for "en"
```
...
 "data": {
    "localeID": "en",
    "language": "en",
    "region": "US",
    "categories": {
      "numbers": {
        "numberSymbols": {
....
```

Issue #148, opened 2019-08-29
[Singleton Service][Get pattern by language&region] plurals rule should always follow with language


nannany00:
Singleton Service API:
GET  /i18n/api/v2/formatting/pattern  (Get pattern from CLDR)

This API is used to provide the pattern data for dates/numbers/currencies/plurals/measurements from CLDR according to the requested language and region.

Plurals rule is related to the translation content, it should always follow with language, as Singleton provides translation by language.

--------------
This issue is:
When the combined locale of language and region is valid, the plurals rule is following the combined locale, not language.

Example:
language=pt; region=PT; scope=plurals

Actual Result is:
it will return the plurals rule of **pt-PT**.

Expected Result is:
it should return the plurals rule of **pt**.

nannany00:
Based on current code, still has issue here:
When language=**pt-PT**, region=**PT**, the plurals rule should be **pt-PT**, not **pt**.

Please help to check.

lyiyu66:
Plurals rule issue has been fixed on current code.

PR #169, opened 2019-09-03
plurals rule should always follow with language

chaoq0318:
https://github.com/vmware/singleton/issues/148: [Singleton Service][Get pattern by language&region] plurals rule should always follow with language
https://github.com/vmware/singleton/issues/149: [Singleton Service][Get pattern by locale] return default region when input is a language

### DIFF
--- a/g11n-ws/modules/md-service-i18n-l2/src/main/java/com/vmware/i18n/l2/service/pattern/PatternServiceImpl.java
+++ b/g11n-ws/modules/md-service-i18n-l2/src/main/java/com/vmware/i18n/l2/service/pattern/PatternServiceImpl.java
@@ -117,7 +117,6 @@ public class PatternServiceImpl implements IPatternService {
 			patternJson = patternDao.getPattern(locale, null);
 			if (StringUtils.isEmpty(patternJson)) {
 				logger.info("file data don't exist");
-				resultData.setPluralSearch(true);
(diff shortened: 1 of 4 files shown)


chaoq0318:
> Why do you update i18n-patterns-core-0.10.31.jar? Is this releated to this issue?
> If not, please remove.

@Xiaochao8 I modified the file in the tool-cldr-extractor model, so I need to update i18n-patterns-core-0.10.31.jar

g11n-ws/modules/md-service-i18n-l2/src/main/java/com/vmware/i18n/l2/service/pattern/PatternServiceImpl.java
```diff
@@ -74,6 +75,13 @@
 		}
 		logger.info("get pattern data from cache");
 		patternMap = JSONUtils.getMapFromJson(patternJson);
+		if (StringUtils.isEmpty(patternMap.get(ConstantsKeys.REGION))) {
+			String regionJson = PatternUtil.getRegionFromLib(locale.replace("_", "-"));
+			if (!StringUtils.isEmpty(regionJson)) {
```
Kai-kevin: It will be better to use isNotEmpty method.

chaoq0318: fixed

Issue #102, opened 2019-08-13
the Etag in the response header keeps on changing for the same GET request


gong-yu:
The timestamp in the response is in all of the API endpoints. Best practice is to have the exact same response body for the exact same GET request. Therefore, timestamps should not be part of any request body. This allows etag headers to have the same value, and browser/clients to be able to cache.

PR #171, opened 2019-09-03
https://github.com/vmware/singleton/issues/102 Etag in the response h…

jessiejuachon:
…eader keeps on changing for the same GET request


jessiejuachon:
> Please sign your commit.

Generated a new gpg signature. It is now shoing "Verified" in the Commit tab.

PR #188, opened 2019-09-16
Correct l10n data attribute naming.

nikichak:
The attribute that holds the key for the translation is data attribute so it should follow the convention.

### DIFF
--- a/samples/client/js/package.json
+++ b/samples/client/js/package.json
@@ -10,7 +10,7 @@
     "load-translation": "load-translation --directory `pwd`/src/sources --product CoreSDKClient --component ui --host http://localhost:8091 --version 1.0 --languages zh,en"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


samples/client/js/package.json
```diff
@@ -10,7 +10,7 @@
     "load-translation": "load-translation --directory `pwd`/src/sources --product CoreSDKClient --component ui --host http://localhost:8091 --version 1.0 --languages zh,en"
   },
   "dependencies": {
-    "@vip/vip-core-sdk": "file:../../vip-vip-core-sdk-0.1.0.tgz",
+    "@vip/vip-core-sdk": "file:../../../vip-vip-core-sdk-0.1.0.tgz",
```
pengrongbo: Why introduce this change here?

nikichak: Actually it is fixed here: https://github.com/vmware/singleton/pull/186

PR #201, opened 2019-09-20
update the S3 configuration file item

tigershi:


jin-jasper:
The values of keys are now meaningless.

Issue #205, opened 2019-09-24
[BUG] [Singleton Service]Fall back to previous version(7.x.x) but not the existing 8.0 when requesting nonexistent 8.0.0.


lyiyu66:
**Describe the bug**
Fall back to previous version(7.x.x) but not the existing 8.0 when requesting nonexistent 8.0.0.
Same behavior when requesting 6.0, it falls back to 5.x.x even though there has 6.0.0.

**To Reproduce**
1. There are the translation bundles of 8.0, 7.9.8, 6.0.0 and 5.9.8.
2. Request the translation of 8.0.0 or 6.0 (when user does not clearly remember the version number).

**Expected behavior**
Should fall back to the closest previous version if the difference of the requested and existing versions is the number of digit is different and value is 0. 
Example:
Request 8.0.0/8 -> 8.0
Request 6.0/6 -> 6.0.0
Request 5.6.0 -> 5.6
Request 7.2 -> 7.2.0

linr211:
not reproduced in my dev environment. Could you double confirm it?

Issue #197, opened 2019-09-20
[BUG][Singleton service]Fall back to 9.6 when requesting nonexistent version(9.6.2)'s translation, even though there has 9.6.1.


lyiyu66:
**Describe the bug**
Majior.Minor.Patch kind version can't fall back to previous same kind version if there is Majior.Minor kind version.
Example: there are versions 9.6, 9.6.1, now request the translation of version 9.6.2, it falls back to 9.6.

**To Reproduce**
1. Translation bundle are existed for version 9.6 and 9.6.1, no 9.6.2 translation.
2. Request the translation of version 9.6.2.

**Expected behavior**
Translation version should fall back to 9.6.1 when requesting 9.6.2.

linr211:
not reproduced in my dev environment, please double check.

lyiyu66:
It is not fixed with latest code, please have a check again.

lyiyu66:
Checked with latest build 445, it is fixed. Hence closing.
![2019-10-22-fixed](https://user-images.githubusercontent.com/51813046/67266385-b819de00-f4e2-11e9-9bc7-9ee797bb7aaa.jpg)

PR #232, opened 2019-10-16
fix issue: https://github.com/vmware/singleton/issues/197

linr211:
fix issue: https://github.com/vmware/singleton/issues/205


g11n-ws/modules/md-restful-api-i18n/src/main/java/com/vmware/vip/i18n/api/base/TranslationProductComponentAction.java
```diff
@@ -201,8 +201,14 @@ private String getMatchedVersion(final String productName, final String version)
 				}
 				for(String s : versionList) {
 					String f = filterVersion(s, version);
-					String ss = StringUtils.isEmpty(matchedVersion) ? s : filterVersion(s, matchedVersion);
-					if((compare(f, version) == -1 || (compare(f, version) == 0 && s.length() < version.length())) && compare(ss, matchedVersion) == 1) {
+					String ss = s;
+					String m = matchedVersion;
+					if(!StringUtils.isEmpty(matchedVersion)) {
```
Xiaochao8: Move filterVersion into compare function, because here needs only compare result.
One suggestion: Assign the first version in the list to matchedVersion, then start comparison from second version.
This will make code and logic clearer.

g11n-ws/modules/md-restful-api-i18n/src/main/java/com/vmware/vip/i18n/api/base/TranslationProductComponentAction.java
```diff
@@ -201,8 +201,14 @@ private String getMatchedVersion(final String productName, final String version)
...
+						if(ss.split("\\.").length > matchedVersion.split("\\.").length) {
+							m = filterVersion(matchedVersion, ss);
+						}
+					}
+					if((compare(f, version) == -1 || (compare(f, version) == 0 && s.length() < version.length())) && compare(ss, m) == 1) {
```
huihuiw01: It's a little difficult to understand this snippet's logic since 6 variables(version, s, f, ss, m, matchedVersion ) are referred to and their values are always changed, is it possible to refactor to make it clear or add comment to explain what each variable represents? especially the meaning of f, ss, m.

It's much better add comment for line 211 since the condition is a little complex.

g11n-ws/modules/md-restful-api-i18n/src/main/java/com/vmware/vip/i18n/api/base/TranslationProductComponentAction.java
```diff
@@ -225,19 +223,31 @@ private int compare(final String source, final String target) {
...
 					break;
 				} else if(Integer.parseInt(s[i]) < Integer.parseInt(t[i])) {
+					b = -1;
 					break;
 				}
 			}
 		}
+
+		if(b == 0) {
+			if(source.length() < target.length()) {
```
Xiaochao8: "9.00.0", "9.0.00" are treated equal. Is this correct?

linr211: I think they should be treated as equal, since if we we compare them part by part, 0 should be same as 00.
"2.0.1" is greater than "2.00000" by the compare function.

PR #302, opened 2019-11-21
Add shared component support in multcomponents

Xiaochao8:


src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -266,19 +264,23 @@ public boolean postString(final Locale locale, final String component,
...
-            logger.error(ConstantsMsg.WRONG_PARAMETER + "locales: {}, components: {}.", locales, components);
+            this.logger.error(ConstantsMsg.WRONG_PARAMETER + "locales: {}, components: {}.", locales, components);
             return retMap;
         }
 
         try {
-            final ComponentsService cs = new ComponentsService(components, locales);
-            retMap = cs.getTranslation();
+            VIPCfg config = this.cfg;
+            if (null != config) {
```
huihuiw01: Should this be  if (null == config) ?

PR #319, opened 2019-12-03
Unit Test for parser.ts/loader.ts/plurals.func.ts

songjiang320:


test/parser.spec.ts
```diff
@@ -0,0 +1,24 @@
+import {defaultResponseParser} from '../src/parser';
+
+describe('Parser Module', () => {
+    it("validateResponse",() => {
+      expect(defaultResponseParser.validateResponse({data:{}})).toEqual(null);
+      expect(defaultResponseParser.validateResponse({ response: { code: 200 }, data: { key: "value" } })).toEqual({ key: "value" });
```
wangyi-e: Do you consider extracting the public part of the response body?

songjiang320: already updated

PR #337, opened 2019-12-18
The pseudo tag is repeated depends on the calls counts with translation-product-api

tigershi:


g11n-ws/modules/md-service-i18n-l3/src/main/java/com/vmware/vip/core/messages/utils/PseudoMessagesUtils.java
```diff
@@ -47,23 +48,22 @@ public static SingleComponentDTO getPseudoMessages(
...
-								pseudoConfig.getExistSourceTag()));
-				pseudoList.add(jo);
-			}
-		}
-		t.setBundles(pseudoList);
-		return t;
-	}
+    public static TranslationDTO getPseudoMessages2(
+            TranslationDTO translationDTO, PseudoConfig pseudoConfig) {
+        TranslationDTO t = SerializationUtils.clone(translationDTO);
```
Xiaochao8: BeanUtils.copyProperties(translationDTO, t); 
Why doesn't this work?

tigershi: It is not a deep copy, so when add the pseudo tag, it will change cache latest source translation

PR #348, opened 2019-12-24
[Singleton-js-client]multiple component support

wangyi-e:


src/services/core.service.ts
```diff
@@ -159,25 +173,83 @@ export class CoreService {
...
     /**
      * return promise with processed result.
      * @param language
      */
-    public loadTranslations(language: string): Promise<any> {
-        const cache = this.cacheManager.lookforTranslationByComponent(this.getComponent(), language);
+    public loadTranslations(language: string): Promise<any> | Promise<any>[] {
+        const isMultipleComponents = this.config.components && this.config.components.length > 0;
+        const isCombine = isMultipleComponents && this.config.combineRequest;
+
```
songjiang320: whether need to logic judge for isMultipleComponents = false and combineRequest = ture when wrong configuration during initilizing VIP

PR #358, opened 2020-01-07
update service jar package file name to singleton-xxxx.jar

surite:


surite:
Must checkin first, CI will pass once another PR https://github.com/vmware/singleton/pull/312 checked in.

PR #413, opened 2020-03-05
add doc of response codes

linr211:


linr211:
> I didn't see any doc of response code, did I?

a pdf file.

linr211:
> PDF file has no preview like readme file.

It should be fine to read on github after checkin.

PR #416, opened 2020-03-06
change log levle to 'trace' 

linr211:


jin-jasper:
it makes log level lower to avoid surplus log data.

PR #425, opened 2020-03-11
G11n java client

jessiejuachon:
1. Adding this API to load configuration from a JSON configuration file. 
The JSON schema shall be common across all supported client libraries.
The content of the sample config file will be enhanced later. 

2. Have also pushed changes for adding a map of cache properties (e.g. etag, cache-control). 
Marked the services (LocaleService, ComponentService, etc) with TODO to indicate unimplemented fetching of these cache properties from the http response.


src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -117,17 +125,46 @@ public void initialize(String vipServer, String productName, String version) {
...
+    public void loadConfig(String configFile) throws VIPClientInitException {
+    	ClassLoader classloader = Thread.currentThread().getContextClassLoader();
+    	InputStream is = classloader.getResourceAsStream(configFile + ".json");
+    	
+    	Reader reader = new InputStreamReader(is);
+    	JSONParser parser = new JSONParser();
+		JSONObject jsonObject = null;
+		try {
+			jsonObject = (JSONObject) parser.parse(reader);
+			this.vipServer = (String) jsonObject.get("online_service_url");
```
linr211: to make it compatible, should it require to do same parsing for properties file's initialization?

jessiejuachon: Support for backwards compatibility:

The new VIPCfg.loadConfig(String cfg) that will load the configuration from a JSON file will be available. However, when an existing application upgrades to the new version of the library, it doesn't have to use this new method. The old VIPCfg.initialize(String cfg) should still work, just marked with @Deprecated. 

Both methods will be available for quite some time until the deprecated method is no longer supported in some future version. In other words, we will remove VIPCfg.initialize(String cfg) eventually, but not immediately.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -117,17 +125,46 @@ public void initialize(String vipServer, String productName, String version) {
...
+    /**
+     * Load client configuration from a JSON file in the resource folder 
+     * 
+     * @param configFile This is the name of the JSON configuration file 
+     * @throws ParseException 
+     * @throws IOException 
+     */
+    public void loadConfig(String configFile) throws VIPClientInitException {
+    	ClassLoader classloader = Thread.currentThread().getContextClassLoader();
+    	InputStream is = classloader.getResourceAsStream(configFile + ".json");
```
linr211: why configFile doesn't contain '.json'?

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
linr211: should validate 'online_service_url'? other parameters as well?

linr211: fine.

sample-client-app/src/main/resources/sampleconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "product": "JavaSample",
+  "version": "1.0.0",
```
linr211: 'product' and 'version' should be removed? since they are defined in 'online_service_url'.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: Makes sense. Will change it.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: Yes, we should do validation for the paramaters. I will add it.
 
Yes, it should read other parameters as well. Let's add them one by one, as we start using them. For example, I am about to add the "prod_mode" parameter because I need to use it for the exception/messaging logic.

sample-client-app/src/main/resources/sampleconfig.json
jessiejuachon: I will remove these 2 properties once the online_service_url has the correct value. I am waiting because there is an ongoing effort to improve API endpoint URLS in the service, right?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentsService.java
```diff
@@ -80,28 +81,37 @@ public ComponentsService(final VIPCfg config) {
...
+	            // TODO pass map of cache properties such as etag and cache control headers
+	            Map<String, Object> cacheProps = null;         
+	            new CacheService(dto).addCacheOfComponent(messages, cacheProps);
+
+	            // update map to return.
+	            dataMap.get(locale).put(comp, messages);
+	        }
+		} catch (IOException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
```
Xiaochao8: Should process this exception?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -31,17 +33,24 @@ public ComponentService(MessagesDTO dto) {
...
      * 
      * @see
      * com.vmware.vipclient.i18n.messages.service.IComponentService#getMessages
      * (com.vmware.vipclient.i18n.base.DataSourceEnum)
      */
     @SuppressWarnings("unchecked")
-    public Map<String, String> getMessages() {
+    public Map<String, String> getMessages(final Map<String, Object> cacheProps) {
         Map<String, String> transMap = new HashMap<String, String>();
+        ComponentBasedOpt cbo = new ComponentBasedOpt(dto);
```
Xiaochao8: Do you think opt is more clear?

src/main/java/com/vmware/vipclient/i18n/base/cache/Cache.java
```diff
@@ -9,22 +9,33 @@
 
 public interface Cache {
 
+	public final static String MESSAGES = "messages";
+	public final static String CACHE_PROPERTIES = "cache_properties";
```
linr211: Move them to ConstantsKeys.java.

src/main/java/com/vmware/vipclient/i18n/base/cache/FormattingCache.java
```diff
@@ -19,12 +20,16 @@ public FormattingCache() {
...
-        return cachedObject == null ? null : (Map<String, String>) cachedObject;
+        Map<String, Object> cache = new HashMap<String, Object>();
+    	if (cachedObject != null) {
+    		cache.put(MESSAGES,  (Map<String, String>) cachedObject);
+    	}
+        return cache;
     }
 
-    public synchronized boolean put(String cacheKey, Map<String, String> map) {
+    public synchronized boolean put(String cacheKey, Map<String, String> map, Map<String, Object> cacheProps) {
```
linr211: no use to this new parameter 'cacheProps'?

src/main/java/com/vmware/vipclient/i18n/base/cache/FormattingCache.java
```diff
@@ -82,6 +87,9 @@ public synchronized int getXCapacity() {
         return this.capacityX;
     }
 
+    public boolean isExpired(String cacheKey) {
```
linr211: Isn't the parameter 'cachkey' used?

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCache.java
```diff
@@ -4,18 +4,27 @@
...
 public class MessageCache implements Cache {
     private String                           id                  = "cache-default";
 
     private long                             expiredTime         = 864000000;                                       // 240hr
     private long                             lastClean           = System.currentTimeMillis();
 
-    private Map<String, Map<String, String>> cachedComponentsMap = new LinkedHashMap<String, Map<String, String>>();
-
+    private final Map<String, Map<String, String>> cachedComponentsMap = new LinkedHashMap<String, Map<String, String>>();
+    private final Map<String, Map<String, Object>> cacheProperties = new LinkedHashMap<String, Map<String, Object>>();
```
linr211: I think it's better to have another container ( or a class similar to MessageCache) to maintain the data of 'cacheProperties', it can find etag and expiration from the container by cachekey.

Basically, MessageCache is for maintaining cached data pool and cacheProperties is for maintaining the etag-expiration. For any searching data from MessageCache, it will firstly detect if expired from  cacheProperties: if yes it should get the etag and make request to remote service, if not it should return cached data from MessageCache.

linr211: I will wait for finalizing this to provide more comments.

src/main/java/com/vmware/vipclient/i18n/base/HttpRequester.java
```diff
@@ -60,7 +61,17 @@ public void setBaseURL(String baseURL) {
     private Map<String, String> customizedHeaderParams = null;
```
Xiaochao8: Change this to static and add another instance variable to store cache control header params?

jessiejuachon: Actually, header request parameters may vary per request. This should be a local variable in the request method and must be set per request. I did not want to change this for now because I wasn't sure why this has been an instance variable. Are we using this setCustomizedHeaderParams anywhere before?

sample-client-app/src/main/resources/sampleconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
jessiejuachon: I am removing it for now. It is not yet being used.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: What does "opt" mean in ComponentBasedOpt?

src/main/java/com/vmware/vipclient/i18n/messages/api/url/URLUtils.java
```diff
@@ -51,4 +66,34 @@ public static boolean isStringInListIgnoreCase(List<String> list,
...
+        	if (responseHeaders != null) {
+	        	List<String> etags = (List<String>) responseHeaders.get(ETAG);
+	        	if (etags != null) {
+	        		String ifNoneMatch = createIfNoneMatchValue(etags);
+	        		Map<String, String> headers = new HashMap<String, String>();
+	        		headers.put(IF_NONE_MATCH_HEADER,ifNoneMatch);
+	        		requester.setCustomizedHeaderParams(headers);
+	        	}
+        	}
+        } else {
```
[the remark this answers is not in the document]
jessiejuachon: The method's return type is 'void' so there should be no return statement. The removeCustomizedHeaderParams in the else statement is for removing any previously added headers. Why are the header parameters stored in an instance variable?

src/main/java/com/vmware/vipclient/i18n/messages/api/url/URLUtils.java
```diff
@@ -51,4 +66,34 @@ public static boolean isStringInListIgnoreCase(List<String> list,
...
+    	if(etags == null || etags.isEmpty()) {
+            return null;
+        }
+        final StringBuilder b = new StringBuilder();
+        final Iterator<String> it = etags.iterator();
+        b.append(it.next());
+        while(it.hasNext()) {
+            b.append(", ").append(it.next());
+        }
+        return b.toString();
```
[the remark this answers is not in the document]
jessiejuachon: String.join is not supported before Java 8

src/main/java/com/vmware/vipclient/i18n/messages/api/url/URLUtils.java
```diff
@@ -51,4 +66,34 @@ public static boolean isStringInListIgnoreCase(List<String> list,
         }
         return false;
     }
+    
+    public static void addIfNoneMatchHeader(Map<String, Object> cacheProps, final HttpRequester requester) {
+    	if (cacheProps != null && !cacheProps.isEmpty()) {
+        	Map<String, List<String>> responseHeaders = (Map<String, List<String>>) cacheProps.get(HEADERS);
+        	if (responseHeaders != null) {
+	        	List<String> etags = (List<String>) responseHeaders.get(ETAG);
+	        	if (etags != null) {
```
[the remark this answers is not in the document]
jessiejuachon: Why? That is the opposite. I want eTags to not be null.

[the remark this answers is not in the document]
jessiejuachon: Actually, this check is not needed anymore because it is already inside createIfNoneMatchValue. I will remove this.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -40,24 +46,25 @@ public CacheService(MessagesDTO dto) {
...
-            Map<String, String> oldmap = c.get(cacheKey);
+            Map<String, Object> oldmap = c.get(cacheKey);
             if (oldmap == null) {
-                c.put(cacheKey, dataMap);
+                c.put(cacheKey, dataMap, cacheProps);
             } else {
                 oldmap.putAll(dataMap);
-                c.put(cacheKey, oldmap);
+                Map<String, String> cachedMessages = (Map<String, String>) oldmap.get(Cache.MESSAGES);
+                c.put(cacheKey, cachedMessages, cacheProps);
```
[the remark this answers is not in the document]
jessiejuachon: Actually, I did not change this logic. It has been there since before. I am also wondering why we need to retain the  old values from the oldmap. In any case, I don't want to change the logic in this PR because it is out of this PR's scope.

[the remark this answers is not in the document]
jessiejuachon: Let's revisit this later.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: Anyway, I changed it to opt.

src/main/java/com/vmware/vipclient/i18n/messages/api/url/URLUtils.java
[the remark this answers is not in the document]
Xiaochao8: There are 2 cases, added or not added.  returning in line after line 79 means added, other cases mean not added.

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCache.java
```diff
@@ -4,18 +4,27 @@
...
```
tigershi, line 22: If you don't need frequent delete or add data, I am not recommended use LinkedHashMap

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -31,41 +32,48 @@ public ComponentService(MessagesDTO dto) {
...
-            Object o = this.getMessages();
+        if (cachedMessages.isEmpty() && !cs.isContainComponent()) {
+        	Map<String, Object> cacheProps = new HashMap<String, Object>();
+            Object o = this.getMessages(cacheProps);
             Map<String, String> dataMap = (o == null ? null
                     : (Map<String, String>) o);
-            cs.addCacheOfComponent(dataMap);
-            retMap = dataMap;
+            
+            cs.addCacheOfComponent(new CacheItem (dataMap, cacheProps));
```
linr211, line 73: Let's say there's a product having 100 components, how many 'cacheProps' instances would be in the memory?

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCache.java
```diff
@@ -184,7 +182,7 @@ public int getCachedKeySize() {
         for (String key : s) {
         	CacheItem cacheItem = this.getCachedTranslationMap().get(key);
             if (cacheItem != null) {              
-                size = size + cacheItem.getCachedData().keySet().size();
+                size += cacheItem.getCachedData().keySet().size();
```
Xiaochao8, line 185: calling KeySet is unnecessary.

src/main/java/com/vmware/vipclient/i18n/messages/service/StringService.java
```diff
@@ -16,37 +16,64 @@
...
+    		// Create a new HashMap to store cache properties.
+    		cacheOfComponent = populateCache(new HashMap<String, Object>(), cacheService, dto);
+       } 
+       return (cacheOfComponent == null || cacheOfComponent.get(key) == null ? "" : cacheOfComponent.get(key));
+    }
+    
+    private volatile boolean running = true;
+	private void populateCacheTask(Map<String, Object> cacheProps, final CacheService cacheService, MessagesDTO dto) {
+		Runnable task = () -> {
+	    	while (running) {
```
Xiaochao8, line 51: It seems this can run only once because running ==false after the first run.

src/main/java/com/vmware/vipclient/i18n/messages/service/StringService.java
```diff
@@ -16,37 +16,64 @@
...
+	}
+	
+	private Map<String, String> populateCache(Map<String, Object> cacheProps, 
+			CacheService cacheService, MessagesDTO dto) {
+    	// Pass cacheProps to getMessages so that:
+		// 1. A previously stored ETag, if any, can be used for the next HTTP request.
+		// 2. Cached properties can be refreshed with new properties from the next HTTP response.
+		Object o = new ComponentService(dto).getMessages(cacheProps);
+		
+		Map<String, String>  cacheOfComponent = (Map<String, String>) o;
```
Xiaochao8, line 71: Suggest changing to Map<String, String>  cacheOfComponent = new ComponentService(dto).getMessages(cacheProps);

src/main/java/com/vmware/vipclient/i18n/messages/service/StringService.java
```diff
@@ -16,37 +16,65 @@
...
+    		// Create a new HashMap to store cache properties.
+    		cacheOfComponent = populateCache(new HashMap<String, Object>(), cacheService, dto);
+       } 
+       return (cacheOfComponent == null || cacheOfComponent.get(key) == null ? "" : cacheOfComponent.get(key));
+    }
+    
+	private void populateCacheTask(Map<String, Object> cacheProps, final CacheService cacheService, MessagesDTO dto) {
+		Runnable task = () -> {
+			boolean running = true;
+	    	while (running) {
```
Xiaochao8, line 51: 'While' loop runs only once because running becomes false after the first round. Do you think a normal function call can do the task?

src/main/java/com/vmware/vipclient/i18n/base/cache/Cache.java
```diff
@@ -110,5 +119,62 @@
...
+    	 * This map is optional and will not be instantiated if not needed.
+    	 */
+    	private Map<String, Object> cacheProperties;
+        
+        public Map<String, String> getCachedData() {
+			return cachedData;
+		}
+        
+        public void addCachedData(Map<String, String> cachedData) {
+			if (cachedData != null) {
```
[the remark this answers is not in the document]
Xiaochao8: It seems I'm wrong, this.cachedData can't be null, but parameter 'cachedData' may be null. The judge is needed still.

src/main/java/com/vmware/vipclient/i18n/base/cache/Cache.java
```diff
@@ -4,27 +4,36 @@
  */
 package com.vmware.vipclient.i18n.base.cache;
```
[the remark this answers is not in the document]
Xiaochao8: The old cache expiration mechanism conflicts with it in the new design. When the CacheExpiredTime is set, it will clear all the cache when it's time.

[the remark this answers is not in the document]
jessiejuachon: It doesn't conflict. It is just different. So we allow both:
1. If CacheExpiredTime is set in vipconfig.properties, then the client library will use the "old" caching like you said, it will clear the entire cache when it is expired. It will not use the max-age from the response.
2. If CacheExpiredTime is not set, then the client library will use the "new" caching and use max-age if it is in the http response. If the http response does not have max-age, then the default 864000000 of MessageCache.expiredTime will be used.

There is no conflict. The product application will use either 1 or 2. The product application will not suddenly break when client library is upgraded.

src/main/java/com/vmware/vipclient/i18n/messages/service/StringService.java
jessiejuachon: No. The task is done in parallel in a separate thread so that the getString method happening in the main thread returns without waiting for the task to finish. This is by design. 

The while loop just makes sure uses the "running=false" in the finally block so that if any runtime exception happens,. the thread is not left alive.. which may lead to memory leak.

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCache.java
```diff
@@ -43,13 +46,35 @@ public void setId(String id) {
...
         if (i != null) {
             hitMap.put(cacheKey, i.intValue() + 1);
         }
-        Object cachedObject = cachedComponentsMap.get(cacheKey);
-        return cachedObject == null ? null : (Map<String, String>) cachedObject;
+        return cachedComponentsMap.get(cacheKey);
+    }
+    
+    public boolean isExpired(String cacheKey) {
+    	CacheItem cacheItem = cachedComponentsMap.get(cacheKey);
```
Xiaochao8, line 58: Do you think it's better to add this at the beginning of the method to make sure that the 2 mechanisms aren't conflicted?
if (VIPCfg.getInstance().getCacheExpiredTime() != VIPCfg.cacheExpiredTimeNotSet) {
    		return false;
}

src/main/java/com/vmware/vipclient/i18n/base/HttpRequester.java
```diff
@@ -50,17 +51,23 @@
...
+     * HTTP headers that are common to all HTTP requests
      */
-    private Map<String, String> customizedHeaderParams = null;
+    private Map<String, String> commonHeaderParams = null;
 
     public void setCustomizedHeaderParams(Map<String, String> params) {
-        customizedHeaderParams = params;
+    	if (params!=null && !params.isEmpty()) {
+    		if (commonHeaderParams == null) {
+    			commonHeaderParams = new HashMap<String, String>();
```
Xiaochao8, line 62: Why not assign params to commonHeaderParams directly?

src/main/java/com/vmware/vipclient/i18n/base/HttpRequester.java
```diff
@@ -50,17 +51,23 @@
...
+    		}
+    		commonHeaderParams.putAll(params);
```
Xiaochao8, line 64: What if product wants to remove some params?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
linr211: I think you didn't get my points.
Firstly it doesn't matter with the 'pre-OOP' at all. It's just a concept for 'container' which may be just a simple hashmap to store the etag with component-key, it's separated from MessageCache.
Secondly, looks like you don't get my question about 100 components. Per your change, i think there will be 100 cacheProps(map) instances in the memory which doesn't work if there are more components, in other words it limits the support to large components case.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -13,6 +13,7 @@
```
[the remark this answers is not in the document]
jessiejuachon: That will be in a separate PR. There is ongoing discussion on whether to support this or not.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: Colin, I think you are the one not getting my point.

See if this helps:
Each CacheProps map has the following key value pairs.
"RESPONSE_TIME" : "xxxxxx"
"HEADER" : "list of headers"
We  store this map inside CacheItem.

Now if you store this in one "container"... let's say a huge HashMap,
the key will still be the cacheKey, and the value will be this map above. You will still store 100 maps, each map associated to the key. The only difference is your putting the cacheProps map in another big hashmap "container" instead of individually, in each CacheItem. Do you see my point now?

src/main/java/com/vmware/vipclient/i18n/base/cache/Cache.java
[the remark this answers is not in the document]
jessiejuachon: This is because the existing code returns null value if http code fails, and actually saves the null value in cache. I agree with you, this is wrong! But I do not want to change the return of the old code because it is our of scope of this PR. But in this PR, I do not want to store null values in the cache so for now, put an empty map there that could be populated later.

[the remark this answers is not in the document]
jessiejuachon: Any suggestions?

src/main/java/com/vmware/vipclient/i18n/base/HttpRequester.java
jessiejuachon: This is out of scope of this PR. Let's not touch it.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: If you still doubt , I have tested implementations with 100 cache keys, 1000 messages each. I compared both implementation's memory consumption. See that using a separate container uses slightly higher heap memory (perhaps due to references)


Using Separate Container class for cacheProps
Using CacheItem cacheProps
Test Code

src/main/java/com/vmware/vipclient/i18n/base/HttpRequester.java
jessiejuachon: Okay, I will bring this back to the original implementation and assign directly.

src/main/java/com/vmware/vipclient/i18n/base/cache/Cache.java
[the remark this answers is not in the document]
Xiaochao8: You can see this in the usage of putAll
     * @throws NullPointerException if the specified map is null
So don't pass null value to the Map.putAll method.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
linr211: What's the reason to store HEADER, RESPONSE_CODE, RESPONSE_MESSAGE? I think they not required to keep in memory; instead, we only need to store etag and response timestamp? so, why we can't just use one map with component as key, etag-timestamp as value? With this, the map size is the number of components, not really huge.
Overall, the principle should keep the memory concise, maintainable and controllable.

jessiejuachon: We can remove response code and response message, but it is the same concept. the key has to be component-locale, the existing cacheKey in MessageCache. Each item in the data cache has a corresponding set of etag/timestamp... you will store this in a separate Map.. as I did above. Same memory consumption, but 2 steps to store/query in different places, 
Frankly, concatenating etag and timestamp adds processing overhead!

And what happens if you add more properties later on you create another map? Very poor design, Colin, Please review OOP concepts.

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
```diff
@@ -12,11 +12,13 @@ public CacheItem() {
 		
 	}
 	
-	public CacheItem (Map<String, String> dataMap, final Map<String, Object> cacheProps) {
+	public CacheItem (Map<String, String> dataMap, String etag, long timestamp, Long maxAgeMillis) {
 		super();
 		if (dataMap != null)
```
Xiaochao8, line 17: Suggest to move this judge into addCachedData method, because It's necessary all the time.

Xiaochao8, line 17: Suggest to move this judge into addCachedData method, because It's necessary all the time.

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
```diff
@@ -0,0 +1,74 @@
...
+	public CacheItem (Map<String, String> dataMap) {
+		super();
+		if (dataMap != null)
+			this.addCachedData(dataMap);
+	}
+	
+	private final Map<String, String> cachedData = new HashMap<String, String>();
+	private String etag;
+	private long timestamp;
+	private Long maxAgeMillis;
```
linr211, line 33: should move maxAgeMillis to MessageCache because the value is same for different CacheItem instance.

src/main/java/com/vmware/vipclient/i18n/base/cache/Cache.java
```diff
@@ -4,27 +4,36 @@
...
 
     /**
-     * put strings to cache by key
+     * check if the cache is expired
      * 
      * @param key
-     * @param map
+     * @return false if expired
+     */
+    public boolean isExpired(String key);
```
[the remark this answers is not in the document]
linr211: I agree with Xiaochao that it's part of cacheitem, and we can still use current isExpired() function for other type of cache implementation.
Reopen it, please double check it.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -47,14 +47,20 @@
     private boolean                    pseudo;
     private boolean                    collectSource;
     private boolean                    cleanCache;
-    private long                       cacheExpiredTime;
+    
+    @Deprecated
```
linr211, line 51: It's not good to add '@Deprecated' fields or functions in the new codes.
How about this judge: if cacheExpiredTime is set, use it to do expiration check and ignore max-age; else use max-age to check?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
linr211: Since there's agreement on reducing cacheProps instances, mark it to resolve.

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCache.java
```diff
@@ -8,8 +8,6 @@
 import java.util.Map;
 import java.util.Set;
 
-import com.vmware.vipclient.i18n.VIPCfg;
-
 public class MessageCache implements Cache {
```
Xiaochao8, line 11: Do you think it's necessary to make this class thread safe?

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -36,29 +37,28 @@ public CacheService(MessagesDTO dto) {
...
-                c.put(cacheKey, dataMap);
-            } else {
-                oldmap.putAll(dataMap);
-                c.put(cacheKey, oldmap);
+            MessageCacheItem cacheItem = (MessageCacheItem) c.get(cacheKey);
+            if (cacheItem == null) {
+            	cacheItem = new MessageCacheItem();
+            	c.put(cacheKey, cacheItem);
             }
+            cacheItem.addCacheItem(itemToCache);
```
Xiaochao8, line 61: How about changing like this? It seems better.
            if (cacheItem == null) {
            	c.put(cacheKey, itemToCache);
            }
            else {
                cacheItem.addCacheItem(itemToCache);
            }

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
```diff
@@ -26,39 +30,65 @@ public ComponentBasedOpt(final MessagesDTO dto) {
...
+        if (response.get(URLUtils.HEADERS) != null)
+        	cacheItem.setEtag(URLUtils.createEtagString((Map<String, List<String>>) response.get(URLUtils.HEADERS)));
+        if (response.get(URLUtils.RESPONSE_TIMESTAMP) != null)
+        	cacheItem.setTimestamp((long) response.get(URLUtils.RESPONSE_TIMESTAMP) );
+        if (response.get(URLUtils.MAX_AGE_MILLIS) != null)
+        	cacheItem.setMaxAgeMillis((Long) response.get(URLUtils.MAX_AGE_MILLIS));
+        Map<String,String> messages = this.getMsgsJson(response);
+        if (messages != null) {
+        	cacheItem.addCachedData(messages);
         }
```
Xiaochao8, line 56: Only when 304 or success, then update cacheItem.

PR #435, opened 2020-03-12
add header check stage

surite:


surite:
Close this PR then open it again, because there are two same travis ci test shown but one does not get result update, that prevent PR merging.

PR #475, opened 2020-03-25
 Improve cache and local bundles

Xiaochao8:


consts.go
```diff
@@ -7,22 +7,29 @@ package sgtn
 
 // api url
 const (
-	i18nApiroot          = "/i18n/api/v2"
```
tigershi: const,  why not use upper case

bundle_dao.go
```diff
@@ -66,8 +84,8 @@ func (d *bundleDAO) getLocales() ([]string, error) {
 	return lSlice, nil
 }
 
-func (d *bundleDAO) getComponentMessages(locale, component string) (ComponentMsgs, error) {
-	compDirPath := filepath.Join(d.cfg.LocalBundles, component)
+func (d *bundleDAO) GetComponentMessages(name, version, locale, component string) (ComponentMsgs, error) {
```
tigershi, line 87: why method return struct，not return the struct object point

Xiaochao8: ComponentMsgs is an interface storing a pointer. So it returns a pointer in fact.

translation_interface.go
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package sgtn
+
+// Translation interface of translation
+type Translation interface {
```
tigershi, line 9: How about change the name Translation to TranslationReader

bundle_dao.go
tigershi: Ok, I have found it

PR #479, opened 2020-03-25
[JsClient] Support downloading both translations and patterns by only…

songjiang320:
… one request to Singleton Service

### DIFF
--- a/src/services/core.service.ts
+++ b/src/services/core.service.ts
@@ -57,6 +57,13 @@ export class CoreService {
         } finally {
             if (callback && typeof callback === 'function') { callback(); }
         }
+        /**
+         * Only when language or region as belew, it returns Promise with {} Object,
+         * on which some products might rely to handle .then() method
+        1. language = 'en-US'
+        2. language = 'en', region = 'US'
+        3. language = 'en' 
+        */
         return Promise.resolve({});
     }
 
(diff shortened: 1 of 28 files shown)


src/services/core.service.ts
```diff
@@ -184,6 +198,35 @@ export class CoreService {
...
+                });
+            }
+        }
+
+        const url = this.getI18nResourceUrl(language, region);
+        const promise = this.coreLoader.getI18nResource(url, this.httpOptions);
+        return promise.then(
+            (result: any) => {
+                const translations = this.resParser.getTranslations(result);
+                if (translations) {
```
wangyi-e, line 218: Logically, translation and pattern data validation should not be nested.

src/services/core.service.ts
```diff
@@ -184,6 +198,35 @@ export class CoreService {
...
+        const cachePattern = this.cacheManager.lookforPattern(language, region);
+        if (typeof cachePattern !== 'undefined') {
+            const cacheTranslation = this.cacheManager.lookforTranslationByComponent(this.config.component, language);
+            if (typeof cacheTranslation === 'undefined') {
+                return Promise.resolve({
+                    components: [cacheTranslation],
+                    pattern: cachePattern
+                });
+            }
+        }
```
wangyi-e, line 211: 1.  If this data can be checked in advance to keep loadTranslationsAndPattern clean.
2. The translation and pattern data validation should not be nested.

songjiang320: Alreay modifed

PR #487, opened 2020-03-27
Improve readme

CiciliaHelena:


README.md
```diff
@@ -36,11 +36,17 @@ Building from source code
...
  git clone git@github.com:vmware/singleton.git
  ```
+ Then go to the root directory.
+ ```
+ cd singleton
+ ```
  Inside singleton/g11n-ws, run a build using Gradle.
  ```
- ./gradlew build
+ g11n-ws/gradlew build
```
nannany00: Which OS do you run this command? Does it work well? The build doesn't be generated on Windows.

CiciliaHelena: Hi, I run this on mac and it works for me. I'm lack of experience with Windows. I take that as the workdir is changed to the root in the previous step. I think you are right the line

> Inside singleton/g11n-ws, run a build using Gradle.

make a confusion. Shall it be better with

> Run a build using Gradle.

The reason I change this is so this line

> java -jar publish/singleton-0.1.0.jar

can be executed without going back upwards to the root again.

surite: I agree, keep working directory consistent, and list all executed command here including command to change directory. So we can make sure user can build Singleton successfully by executing these command one by one.

nannany00: Hi @CiciliaHelena
I agree to change it like you said:
> Run a build using Gradle.

But also change the command to like:
```
cd g11n-ws
gradlew build
```
Same thing
```
cd publish
java -jar singleton-0.1.0.jar
```
Is it better? or any other thinking?

CiciliaHelena: Hi @nannany00 
I just experiment with a dockerfile and you are right, `g11n-ws/gradlew` build somehow can't build the code and require me to go to g11n-ws. I think it's best to do what you have just suggested. I'll make a change. Thanks.

PR #506, opened 2020-04-10
upload the test code for csharp

lyiyu66:


lyiyu66:
Close it as it is missing signature.

PR #511, opened 2020-04-15
Loading offline resource bundles

jessiejuachon:


src/main/java/com/vmware/vipclient/i18n/util/FileUtil.java
```diff
@@ -19,7 +20,31 @@
 
 public class FileUtil {
     static Logger logger = LoggerFactory.getLogger(FileUtil.class);
-
+    
+    public static JSONObject readJson(String pathName)  {
```
linr211, line 24: Does this function work if the bundle is out of the client(jar)?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -4,24 +4,47 @@
...
     }
 
+    @Deprecated
     public JSONObject getComponentMessages() {
         return JSONBundleUtil.getMessages(dto.getLocale(), dto.getProductID(),
                 dto.getVersion(), dto.getComponent());
     }
+    
+    @Override
+    public void getComponentMessages(MessageCacheItem cacheItem) {
```
linr211, line 35: how about using 'refreshComponentMessages' as the function name?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -4,24 +4,47 @@
...
+    	String pathName = VIPCfg.getInstance().getOfflineResourcesBaseUrl();
+    	String filePath = FormatUtils.format(OFFLINE_RESOURCE_PATH, dto.getComponent(), dto.getLocale());
+    	if (pathName != null) {
+    		pathName = pathName.concat(filePath);
+    	} else {
+    		pathName = filePath;
+    	}
+    	Map<String, String> messages = JSONBundleUtil.getMessages(pathName);
```
linr211, line 43: should judge if messages is empty before adding to cache.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -4,24 +4,47 @@
...
         return JSONBundleUtil.getMessages(dto.getLocale(), dto.getProductID(),
                 dto.getVersion(), dto.getComponent());
     }
+    
+    @Override
+    public void getComponentMessages(MessageCacheItem cacheItem) {
+    	String pathName = VIPCfg.getInstance().getOfflineResourcesBaseUrl();
+    	String filePath = FormatUtils.format(OFFLINE_RESOURCE_PATH, dto.getComponent(), dto.getLocale());
+    	if (pathName != null) {
+    		pathName = pathName.concat(filePath);
```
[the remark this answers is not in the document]
Xiaochao8: I meant line 39. 
Wtich Paths.get the, the '/' in offlineResourcesBaseUrl of VIPCfg will have no affect on the result. 
Please check.

Xiaochao8:
By new flow, client will fallback to local bundles if failing to get from server.
I couldn't find changes on this. Could you check this?

jessiejuachon:
> By new flow, client will fallback to local bundles if failing to get from server.
> I couldn't find changes on this. Could you check this?



> By new flow, client will fallback to local bundles if failing to get from server.
> I couldn't find changes on this. Could you check this?

@Xiaochao8 , let's do that in a separate PR.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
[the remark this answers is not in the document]
jessiejuachon: > By new flow, client will fallback to local bundles if failing to get from server.
> I couldn't find changes on this. Could you check this?

Sure.  I was planning to check it in the next Pr so that you are not overwhelmed. I have just checked it in now.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
linr211: but it returns 'void', meaning outside caller get nothing. If there's case to detect the change on service, how to do it?

jessiejuachon: Caller of the method gets the change inside the object that is passed as argument.

"All object references in Java are passed by value. This means that a copy of the value will be passed to a method. Passing a copy of the value also changes the real value of the object. "

What do you mean by "case to detect the change on service, how to do it"?

src/main/java/com/vmware/vipclient/i18n/I18nFactory.java
```diff
@@ -76,10 +76,13 @@ public Message getMessageInstance(Class c, VIPCfg cfg) {
...
-        } else if (this.getCfg().getVipServer() == null
-                || this.getCfg().getProductName() == null
+        } else if (this.getCfg().getProductName() == null
                 || this.getCfg().getVersion() == null) {
-            logger.error("VipServer|ProductName|Version is null!");
+            logger.error("ProductName|Version is null!");
+            return i;
+        } else if (this.getCfg().getVipServer() == null 
+        		&& this.getCfg().getOfflineResourcesBaseUrl() == null) {
+        	logger.error("One of offlineResourcesBaseUrl and vipServer must not be null!");
```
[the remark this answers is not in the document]
jessiejuachon: Yes, I agree. But let's not do it in this PR. The scope of this is for offline mode support only. Let's do the improvements in a separate PR. 
I also do not like that the I18nFactory has its own instance of VIPCfg, but let's not touch this for now.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
```diff
@@ -57,13 +60,27 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
...
-		        }
-        	}
+			      
+	        if (responseCode.equals(HttpURLConnection.HTTP_OK)) {
+		        JSONObject respObj = (JSONObject) JSONValue.parse((String) response.get(URLUtils.BODY));
+		        try {
+	        		if (getResponseCode(respObj) == 200 && 
+	        				// Do not use service response if not matching the locale in the request
+	        				(LocaleUtility.isDefaultLocale(this.dto.getLocale()) || 
+	        						!LocaleUtility.isDefaultLocale(getLocale(respObj)))){
```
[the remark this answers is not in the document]
jessiejuachon: It actually is. Anyway, I have updated it to be more clear.

src/main/java/com/vmware/vipclient/i18n/messages/service/StringService.java
```diff
@@ -22,33 +22,57 @@
...
     	String key = dto.getKey();
+    	
+    	MessageCacheItem cacheItem = getComponentLocaleCache(dto);
+    	
+    	// If failed to get MessageCacheItem of a non-default locale, 
+    	// use MessageCacheItem of the default locale instead. 
+    	Map<String, String> cacheOfComponent = cacheItem.getCachedData();	
+    	if (cacheOfComponent.isEmpty() && !LocaleUtility.isDefaultLocale(dto.getLocale())) {
+			MessagesDTO defaultLocaleDTO = new MessagesDTO(dto.getComponent(), dto.getComment(), 
+					dto.getKey(), dto.getSource(), LocaleUtility.defaultLocale.toLanguageTag(), null);
```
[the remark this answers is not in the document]
jessiejuachon: I am not changing this logic in this PR. The only change is to support offline mode. Let's do it in the next PR.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
```diff
@@ -65,7 +65,10 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
 		        JSONObject respObj = (JSONObject) JSONValue.parse((String) response.get(URLUtils.BODY));
 		        try {
 	        		if (getResponseCode(respObj) == 200 && 
-	        				// Do not use service response if not matching the locale in the request
+	        				// If the response is 200, there are 2 possible cases:
+	        				// 1. the response's locale matches the request's locale OR
+	        				// 2. the response's locale is for the default locale (service's fall back behavior)
```
Xiaochao8, line 70: Service is removing fallback behavior.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
```diff
@@ -65,7 +65,10 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
...
+	        				// Use the response only if its locale matches the request's locale
 	        				(LocaleUtility.isDefaultLocale(this.dto.getLocale()) || 
 	        						!LocaleUtility.isDefaultLocale(getLocale(respObj)))){
```
Xiaochao8, line 73: I don't see where 'match' is. Could you point out?

src/test/java/com/vmware/vipclient/i18n/messages/service/OfflineModeTest.java
```diff
@@ -0,0 +1,196 @@
...
+        dto.setProductID(VIPCfg.getInstance().getProductName());
+        dto.setVersion(VIPCfg.getInstance().getVersion());
+        
+    	CacheService cs = new CacheService(dto);
+    	
+    	translation.getString(locale, component, key, source, comment, args);
+    	
+    	MessageCacheItem cacheItem = cs.getCacheOfComponent();
+    	assertNotNull(cacheItem);
+    	assertEquals(messageFil, cacheItem.cachedData.get(key));
```
linr211, line 74: To get the string, the logic is not coherent, e.g. translation.getString is parallel with cacheItem.cachedData.get(key)
do you think it make sense?
My suggestion is CacheService should be in translation.getString, it should be not exposed to API caller(developer).

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -43,6 +43,7 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
 					getResource(path.toString()).toURI());
 			Map<String, String> messages = JSONBundleUtil.getMessages(path);
 	    	cacheItem.addCachedData(messages);
+	    	cacheItem.setTimestamp(System.currentTimeMillis());
```
Xiaochao8, line 46: Suggest to set 'never expire'

jessiejuachon: It has to expire becasue maybe the service was just temporarily down. So that when it expires, fetch from service can be re-tried.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
Xiaochao8: We don't need to consider fallback because service won't fallback in the future.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
Xiaochao8: When service is empty, it should be set as never expire.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -4,31 +4,58 @@
...
+			Map<String, String> messages = JSONBundleUtil.getMessages(path);
+	    	cacheItem.addCachedData(messages);
+	    	cacheItem.setTimestamp(System.currentTimeMillis());
+		} catch (Exception e) {
+			// Do not update cacheItem
+		}
+    }
 
+    @Override
     public String getString() {
```
Xiaochao8, line 53: We don't need to provide such fine-grained interface internally. Suggest transfering component data internally, then getting string message in the last step.
Cache works in this way.

src/main/java/com/vmware/vipclient/i18n/util/FormatUtils.java
```diff
@@ -30,7 +30,10 @@ public static String format(String pattern, Object... arguments) {
     }
 
     public static String format(String pattern, Locale locale, Object... arguments) {
-        MessageFormat messageFormat = new MessageFormat(pattern, locale);
-        return messageFormat.format(arguments, new StringBuilder(), new FieldPosition(0)).toString();
+    	if (pattern != null && !pattern.isEmpty() && arguments != null && arguments.length > 0) {
```
Xiaochao8, line 33: Should log error when check fails?

src/test/java/com/vmware/vipclient/i18n/messages/service/OfflineModeTest.java
linr211: Are both(transaltion.getString and cacheItem.cachedData.get(key) sequential? What transaltion.getString will return for the first call?

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -48,28 +49,112 @@ public TranslationMessage() {
...
+     * @param locale The locale in which the message is requested to be localized
+     * @param args Values to replace placeholders in the message with
+     * @return One of the items in the following priority-ordered list: 
+     * <ul>
+     * 		<li>The message in the requested locale</li> 
+     * 		<li>The message in the default locale</li>
+     * 		<li>null</li>
+     * </ul>
+     */
+    private String getCachedMessage(String component, String key, Locale locale, Object[] args) {
```
Xiaochao8, line 89: args is unnecessary.

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -48,28 +49,112 @@ public TranslationMessage() {
...
+    		} else if (!VIPCfg.getInstance().isPseudo()) {
+		    	// If the source message is not equal to the cached source (loaded from remote or from offline bundle file),
+		    	// it means that this source message hasn't been collected for localization, so return the source message
+	    		MessagesDTO sourceLocaleDTO = new MessagesDTO(component, key, source, sourceOpt.getLocale().toLanguageTag(), this.cfg);
+	    		StringService s = new StringService();
+	            String cachedSrcLocaleMsg = s.getString(sourceLocaleDTO);
+	            if (!source.equals(cachedSrcLocaleMsg) || 
+	            		cachedSrcLocaleMsg == null || cachedSrcLocaleMsg.isEmpty()) {
+	            	return FormatUtils.format(source, sourceOpt.getLocale(), args);
+	            }
```
Xiaochao8, line 143: Still need to consider if isPseudo == true?

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -48,28 +49,112 @@ public TranslationMessage() {
...
+    			}
+    			return message;
+    		} else if (!VIPCfg.getInstance().isPseudo()) {
+		    	// If the source message is not equal to the cached source (loaded from remote or from offline bundle file),
+		    	// it means that this source message hasn't been collected for localization, so return the source message
+	    		MessagesDTO sourceLocaleDTO = new MessagesDTO(component, key, source, sourceOpt.getLocale().toLanguageTag(), this.cfg);
+	    		StringService s = new StringService();
+	            String cachedSrcLocaleMsg = s.getString(sourceLocaleDTO);
+	            if (!source.equals(cachedSrcLocaleMsg) || 
+	            		cachedSrcLocaleMsg == null || cachedSrcLocaleMsg.isEmpty()) {
```
Xiaochao8, line 141: It seems that this line is unnecessary.

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -48,28 +49,112 @@ public TranslationMessage() {
...
+	            	return FormatUtils.format(source, sourceOpt.getLocale(), args);
+	            }  
+		    	
+    		}
+    	}
+    	return FormatUtils.format(message, locale, args);
```
Xiaochao8, line 147: Need to consider if adding ConstantsKeys.PSEUDOCHAR2?

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -47,29 +48,104 @@ public void setCfg(final VIPCfg cfg) {
...
+    	SourceOpt sourceOpt = VIPCfg.getInstance().getSrcOpt();
+    	String source = (sourceOpt == null) ? null : sourceOpt.getMessage(key);
+    	
+    	// If sourceOpt is defined, pseudo-translation which uses the source message is supported
+    	if (VIPCfg.getInstance().isPseudo() && source != null && sourceOpt.getLocale().equals(locale)) {
+    		return ConstantsKeys.PSEUDOCHAR2 + FormatUtils.format(source, sourceOpt.getLocale(), args) + ConstantsKeys.PSEUDOCHAR2;
+    	}
+    	
+    	// Get the message in the target locale
+    	message = getCachedMessage(component, key, locale);
```
linr211, line 104: if the message is not empty, should return it and skip source handling?

src/main/java/com/vmware/vipclient/i18n/messages/service/StringService.java
```diff
@@ -22,33 +22,59 @@
...
     public String getString(MessagesDTO dto) {
     	String key = dto.getKey();
+    	
+    	MessageCacheItem cacheItem = getComponentLocaleCache(dto);
+    	
+    	// If failed to get MessageCacheItem of a non-default locale, 
+    	// use MessageCacheItem of the default locale instead. 
+    	Map<String, String> cacheOfComponent = cacheItem.getCachedData();	
+    	if (cacheOfComponent.isEmpty() && !LocaleUtility.isDefaultLocale(dto.getLocale())) {
+			MessagesDTO defaultLocaleDTO = new MessagesDTO(dto.getComponent(),
```
linr211, line 40: For the first call, the cacheOfComponent will be empty and then fallback to default value? is this correct?

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -393,4 +436,27 @@ public void setCachePath(String cachePath) {
...
+
+	public void setOfflineResourcesBaseUrl(String offlineResourcesBaseUrl) {
+		this.offlineResourcesBaseUrl = offlineResourcesBaseUrl;
+	}
+
+	public List<DataSourceEnum> getMsgOriginsQueue() {
+		return msgOriginsQueue;
+	}
+
+	public void setMsgOriginsQueue(List<DataSourceEnum> msgOriginsQueue) {
```
linr211, line 451: I don't see how to use this function 'setMsgOriginsQueue' and how to create the parameter 'msgOriginsQueue'.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -56,7 +62,18 @@
...
+    
+    /**
+     * The optional SourceOpt object to access source messages. If SourceOpt is defined, then you can use a source message:
+     * 	a. as fallback if neither localized message nor default locale message was not retrieved successfully
+     * 	b. for pseudo-translation
+     * 	c. if the message hasn't been collected for localization
+     * e.g. ResourceBundleSrcOpt is the SourceOpt implementation that retrieves source messages from a .properties file. 
+     * If source messages need to come from another location such as a DB, then have another implementation like DBSourceOpt.
+     */
+    private SourceOpt				   srcOpt;
```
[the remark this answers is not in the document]
jessiejuachon: Where would you like to have it?

src/main/java/com/vmware/vipclient/i18n/I18nFactory.java
```diff
@@ -53,6 +53,7 @@ public static synchronized I18nFactory getInstance(VIPCfg cfg) {
         if (factory == null) {
             factory = new I18nFactory(cfg);
         }
+        factory.setCfg(cfg);
```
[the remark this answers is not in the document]
jessiejuachon: For testing purposes where different tests may use different config files. See CacheServiceTesrt.java

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -86,6 +103,10 @@ public static synchronized VIPCfg getInstance() {
         }
         return gcInstance;
     }
+    
+    public static synchronized void resetInstance() {
```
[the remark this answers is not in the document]
jessiejuachon: See @Before in CacheServiceTest.java and OfflineModeTest.java
VipCfg is different for each test so have to reset it. 
I see that for other tests, you are manually putting configurations back to original as a workaround.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -120,9 +141,19 @@ public void initialize(String vipServer, String productName, String version) {
     }
     
     /**
-     * initialize the instance by a properties file
+     * Initialize VIPCfg instance using a configuration file
+     * 
+     * @param cfg The configuration file
+     * @param srcOpt The optional SourceOpt object which gives access to source messages
+     */
+    public void initialize(String cfg, SourceOpt srcOpt) throws VIPClientInitException {
```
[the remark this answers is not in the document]
jessiejuachon: Where would you like to define the source? You hardcoded the source type to a ResourceBundle in getString2, which does not address the use case you mentioned the other time where source messages could come from DB or something else. 

What is your solution to this? Using the old getString with source parameter is not acceptable because it is error prone and devs have to type the source manually every time they need to use the string.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -303,10 +342,12 @@ public void setPseudo(boolean pseudo) {
         this.pseudo = pseudo;
     }
 
+    @Deprecated
```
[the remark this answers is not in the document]
jessiejuachon: What would you like me to remove? The @Deprecated annotations?

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: This function is not used, I can remove it if you like.
See VIPCfg.initialize: 
```// Remote VIP resources takes priority over offline resources
        // so add DataSourceEnum.VIP first to msgOriginsQueue
        if (prop.containsKey("vipServer")) {
            this.vipServer = prop.getString("vipServer");
            this.msgOriginsQueue.add(DataSourceEnum.VIP);
        }
        if (prop.containsKey("offlineResourcesBaseUrl")) {
        	this.offlineResourcesBaseUrl = prop.getString("offlineResourcesBaseUrl");
        	this.msgOriginsQueue.add(DataSourceEnum.Bundle);
        }
```

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -393,4 +436,27 @@ public void setCachePath(String cachePath) {
...
+		this.msgOriginsQueue = msgOriginsQueue;
+	}
+
+	public SourceOpt getSrcOpt() {
+		return srcOpt;
+	}
+
+	public void setSrcOpt(SourceOpt srcOpt) {
+		this.srcOpt = srcOpt;
+	}
```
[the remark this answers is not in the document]
jessiejuachon: Where would you like to put it?

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
jessiejuachon: That was my initial thought, but the code in getString uses the source message for the following:
    ```
	// 	a. if neither localized message nor default locale message was not retrieved successfully
    	// 	b. for pseudo-translation
    	//      c. if the message hasn't been collected for localization
```
Can we skip these?

File: src/main/java/com/vmware/vipclient/i18n/messages/service/StringService.java (hunk 19 above)
Comment (jessiejuachon) [reply src/main/java/com/vmware/vipclient/i18n/messages/service/StringService.java:40]: Yes, this is correct. If query for the target locale failed, then query for the default locale (which may or may not be the same as the source locale).  See the workflow design:

https://confluence.eng.vmware.com/display/GQ/Singleton+Client+Library+-+Caching%2C+Fallback+Mechanism%2C+and+Other+Features#SingletonClientLibrary-Caching,FallbackMechanism,andOtherFeatures-L10nQuery

File: src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java [hunk 27]
```diff
@@ -4,31 +4,49 @@
...
-        return v;
+    
+    @Override
+    public void getComponentMessages(MessageCacheItem cacheItem) {
+    	String resource = VIPCfg.getInstance().getOfflineResourcesBaseUrl();
+    	String filePath = FormatUtils.format(OFFLINE_RESOURCE_PATH, dto.getComponent(), dto.getLocale());
+    	Path path = Paths.get(resource, filePath);
+		try {
+			path = Paths.get(Thread.currentThread().getContextClassLoader().
+					getResource(path.toString()).toURI());
```
Xiaochao8, line 43: The path should be relative to current working directory instead of Thread.currentThread().getContextClassLoader(), to make it easy to collect source.
Source collection will make use of config file.

src/main/java/com/vmware/vipclient/i18n/I18nFactory.java
Xiaochao8, line 56: Need to remove this?

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -393,4 +411,24 @@ public void setCachePath(String cachePath) {
...
+
+	public List<DataSourceEnum> getMsgOriginsQueue() {
+		return msgOriginsQueue;
+	}
+
+	public void addMsgOriginsQueue(DataSourceEnum dataSource) {
+		this.msgOriginsQueue.add(dataSource);
+	}
+
+	public void setMsgOriginsQueue(List<DataSourceEnum> msgOriginsQueue) {
```
Xiaochao8, line 430: Remove this 2 methods or make them private?
They don't need to exported.

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
```diff
@@ -31,16 +32,28 @@ public MessageCacheItem (Map<String, String> dataMap) {
...
+		this.cachedData.put(key, value);
+	}
+	
+	public boolean isCachedDataEmpty() {
+		return this.cachedData.isEmpty();
+	}
 	
 	public synchronized void addCachedData(Map<String, String> cachedData) {
 		if (cachedData != null) 
 			this.cachedData.putAll(cachedData);
```
Xiaochao8, line 49: If cachedData == null, then do nothing?

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
```diff
@@ -31,16 +32,28 @@ public MessageCacheItem (Map<String, String> dataMap) {
...
 	}
 	
-	public synchronized void addCacheItem (MessageCacheItem cacheItem) {
+	public synchronized void setCacheItem (MessageCacheItem cacheItem) {
+		// Do not update cacheItem if timestamp is earlier than current. 
+		// An older timestamp comes from an old thread that was blocked.
+		if (cacheItem.getTimestamp() < this.timestamp) 
+			return;
```
Xiaochao8, line 56: In this logic, we should prevent querying fro server parallelly.

Xiaochao8: We can leave it if much effort.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: This is just a standard setter. This can be useful.

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
jessiejuachon: Yes, do nothing because there is nothing to add.

src/main/java/com/vmware/vipclient/i18n/I18nFactory.java
jessiejuachon: Next PR

PR #544, opened 2020-05-08
Error logging for Unsuccessful L10n Fetch ( plus code clean up)

jessiejuachon:


src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
```diff
@@ -52,13 +52,6 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
         
         if (responseCode != null && (responseCode.equals(HttpURLConnection.HTTP_OK) || 
         		responseCode.equals(HttpURLConnection.HTTP_NOT_MODIFIED))) {
-        	
-        	if (response.get(URLUtils.RESPONSE_TIMESTAMP) != null)
-	        	cacheItem.setTimestamp((long) response.get(URLUtils.RESPONSE_TIMESTAMP) );
```
Xiaochao8, line 57: This should be kept here because timestamp need to update when 304.

src/main/java/com/vmware/vipclient/i18n/util/ConstantsKeys.java
```diff
@@ -67,5 +67,5 @@
     public static final String      DISPLAY_NAME                       = "displayName";
     public static final String      JSON_KEYSET                        = "jsonkeyset";
     public static final String      TRANSLATION_STATUS                 = "translationStatus";
-
+    public static final String   	GET_MESSAGES_FAILED				   = "Failed to get messages for component {0}, locale: {1}.";
```
linr211, line 70: should move it to ConstantsMsg.java

PR #556, opened 2020-05-15
Updating the sample application

jessiejuachon:


src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -63,7 +63,7 @@ public JSONObject getPatterns(String language, String region) {
 
     private JSONObject getPatternsFromBundle(String locale) {
         JSONObject patterns = null;
-        if (LocaleUtility.isDefaultLocale(locale)) {
+        if (LocaleUtility.isDefaultLocale(locale) || locale.contains("en")) {
```
linr211, line 66: are you sure locale.contains("en") means the locale is default? A locale string could easily contain 'en'.

jessiejuachon: Rolled back this cha nge

PR #629, opened 2020-06-19
Add capability of loading local formatting pattern files

huihuiw01:


src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -30,63 +33,90 @@ public JSONObject getPatternsByCategory(String locale, String category) {
...
         }
         return patterns;
     }
 
     private JSONObject getPatternsFromBundle(String language, String region) {
         JSONObject patterns = null;
-        if (VIPCfg.getInstance().getMessageOrigin() == DataSourceEnum.VIP) {
-        	patterns = new RemotePatternOpt().getPatternsByLocale(language, region);
+        Locale locale = new Locale(language, region);
+        if (LocaleUtility.isDefaultLocale(locale)) {
```
jessiejuachon, line 103: (Same comment as above)
This does not follow the client workflow design. Always try to fetch from the service first even for the default locale, so remove the if-statement block.

src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -30,63 +33,90 @@ public JSONObject getPatternsByCategory(String locale, String category) {
...
+            logger.debug("Look for pattern from local bundle for language [{}], region [{}]!", language, region);
+            patterns = new LocalPatternOpt()
+                    .getEnPatterns(ConstantsKeys.EN);
         } else {
-            patterns = new LocalPatternOpt().getPatternsByLocale(ConstantsKeys.EN);
+            if (VIPCfg.getInstance().getMsgOriginsQueue().get(0) == DataSourceEnum.VIP) {
```
jessiejuachon, line 108: (Same comment as above)
Do not use if-else statements to get the message origin from the msgOriginsQueue because the order may change in the future. in other words, .get(0) may not always return DataSourceEnum.VIP even if online mode is supported.

Instead, iterate over msgOriginsQueue. See ComponentService.fetchMessages for reference to do this. Add a createPatternOpt method inside DataSourceEnum.Bundle which will return new LocalePatternOpt(). Add the same method inside DataSourceEnum.VIP which will return new RemotePatternOpt(). Use this DataSourceEnum.createPatternOpt just like in ComponentService.fetchMessages.

If you follow the above, the fallback from online to offline mode (and to any other mode added in the queue in the future) will naturally happen without having to modify this method.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalPatternOpt.java
```diff
@@ -4,21 +4,39 @@
...
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+import java.util.Map;
 
 public class LocalPatternOpt {
 
-    public JSONObject getPatternsByLocale(String locale) {
+    public JSONObject getEnPatterns(String locale) {
```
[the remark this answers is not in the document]
huihuiw01: This follows the old workflow, remove this after implementing new workflow.

src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -30,63 +33,90 @@ public JSONObject getPatternsByCategory(String locale, String category) {
...
+        patterns = getPatternsFromBundle(locale);
+        if (patterns != null) {
+            logger.debug("Find the pattern for locale [{}].\n", locale);// [datetime] and
+            new PatternCacheService().addPatterns(locale, patterns);
+            logger.debug("Pattern is cached for locale [{}]!\n\n", locale);
+            return patterns;
+        }
+        if (!LocaleUtility.isDefaultLocale(locale)) {
+            logger.info("Can't find pattern for locale [{}], look for English pattern as fallback!", locale);
+            patterns = getPatterns(ConstantsKeys.EN);
```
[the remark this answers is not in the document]
huihuiw01: This follows the old workflow, change to use default locale after implementing new workflow.

src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -30,63 +33,90 @@ public JSONObject getPatternsByCategory(String locale, String category) {
...
+        if (patterns != null) {
+            logger.debug("Find the pattern for language [{}], region [{}].\n", language, region);// [datetime]
+            // and
+            new PatternCacheService().addPatterns(key, patterns);
+            logger.debug("Pattern is cached for language [{}], region [{}]!\n\n", language, region);
+            return patterns;
+        }
+        if (!LocaleUtility.isDefaultLocale(new Locale(language, region))) {
+            logger.info("Can't find pattern for language [{}] region [{}], look for English pattern as fallback!", language, region);
+            patterns = getPatterns(ConstantsKeys.EN);
```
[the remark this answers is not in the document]
huihuiw01: This follows the old workflow, change to use default locale after implementing new workflow.

src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -4,89 +4,107 @@
...
+            return patterns;
+        }
+        patterns = getPatternsFromDS(locale, VIPCfg.getInstance().getMsgOriginsQueue().listIterator());
+        if (patterns != null) {
+            logger.debug("Find the pattern for locale [{}].\n", locale);// [datetime] and
+            new PatternCacheService().addPatterns(cacheKey, patterns);
+            logger.debug("Pattern is cached for locale [{}]!\n\n", locale);
+            return patterns;
+        }
+        if (!LocaleUtility.isDefaultLocale(locale)) {
```
jessiejuachon, line 50: Put this fallback logic in PatternMessage so that this method here in PatternService is always for the requested language only.

Once you move the fallback logic into PatternMessage, update the cache in there in such a way that the key with the requested language is mapped to the FormatCacheItem of the fallback language.
`new PatternCacheService().addPatterns(<keyOfRequestedLanguage>, <patternsOfDefaultLanguage>);`

Without the above proposed changes, your code will always try to fetch the requested locale first before using the fallback locale that is already in the cache. With the above proposed changes, any succeeding calls for the requested language will immediately return the cached fallback locale until it expires.

Note: You can either put the above logic in PatternMessage as I propose (just like how it is in TranslationMessage.getMessages), or create another method here in PatternService to keep the logic in the "service layer".

src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -4,89 +4,107 @@
...
+        }
+        patterns = getPatternsFromDS(language, region, VIPCfg.getInstance().getMsgOriginsQueue().listIterator());
+        if (patterns != null) {
+            logger.debug("Find the pattern for language [{}], region [{}].\n", language, region);// [datetime]
+            // and
+            new PatternCacheService().addPatterns(key, patterns);
+            logger.debug("Pattern is cached for language [{}], region [{}]!\n\n", language, region);
+            return patterns;
+        }
+        if (!LocaleUtility.isDefaultLocale(new Locale(language, region))) {
```
jessiejuachon, line 80: Same comment as above (line 50)

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalPatternOpt.java
```diff
@@ -4,17 +4,51 @@
...
+        logger.debug("Look for pattern from local bundle for language [{}], region [{}]!", language, region);
+        LocaleDataDTO resultData = CommonUtil.getLocale(language, region);
+        String normalizedLocale = resultData.getLocale();
+        logger.debug("Normalized locale for language [{}], region [{}] is [{}]", language, region, normalizedLocale);
+        return getPatternsByLocale(normalizedLocale);
+    }
 
-    public JSONObject getPatternsByLocale(String locale) {
-        Map<String, Object> patterns = PatternBundleUtil.readJSONFile(locale);
+    public JSONObject getPatternsByLocale(String normalizedLocale) {
```
jessiejuachon, line 42: Can this be private now? If yes, please change it to private.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
```diff
@@ -42,27 +57,98 @@
...
-				return fileName.substring(BUNDLE_PREFIX.length(), fileName.indexOf('.'));
-			}).forEach(s->supportedLocales.put(s, Locale.forLanguageTag(s).getDisplayName(inLocale)));
+				if(fileName.startsWith(BUNDLE_PREFIX) && fileName.endsWith(BUNDLE_SUFFIX)) {
+					return fileName.substring(BUNDLE_PREFIX.length(), fileName.indexOf('.'));
+				}
+				return "";
+			}).forEach(language -> {
+				if(language != null && !"".equalsIgnoreCase(language) && !ConstantsKeys.SOURCE.equalsIgnoreCase(language)) {
+					if (languagesNames != null) {
+						supportedLocales.put(language, (String) languagesNames.get(language));
```
Xiaochao8, line 84: Should call ProductService/LocalProductOpt to get this information rather than getting directly.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
```diff
@@ -42,27 +57,98 @@
...
 			logger.debug(e.getMessage());
 		}
     	return supportedLocales;
     }
-    
-    private void getSupportedLocales(Path path, Map<String, String> supportedLocales, String displayLanguage) throws IOException {
-    	Locale inLocale = Locale.forLanguageTag(displayLanguage); 
+
+    private void getSupportedLocales(Path path, Map<String, String> supportedLocales, JSONObject languagesNames) throws IOException {
     	try (Stream<Path> listOfFiles = Files.walk(path).filter(p -> Files.isRegularFile(p))) {
```
Xiaochao8, line 74:
```suggestion
    	try (Stream<Path> listOfFiles = Files.walk(path).filter(p ->p.toFile().isFile())) {
```

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
```diff
@@ -42,27 +57,98 @@
...
+						supportedLocales.put(language, (String) languagesNames.get(language));
+					} else {
+						supportedLocales.put(language, "");
+					}
+				}
+			});
         }
     }
 
+    private JSONObject getLanguagesNames(String displayLanguage){
```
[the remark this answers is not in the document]
huihuiw01: Yes, because in online mode Singleton service get the language names from CLDR, so for offline mode I change the code to get the language names from CLDR, too.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -30,66 +31,112 @@ public LocaleService() {
...
+            return regionMap;
+        }
+        if (!LocaleUtility.isDefaultLocale(language)) {
+            logger.info("Can't find regions for locale [{}], look for English regions as fallback!", language);
+            regionMap = getTerritoriesFromCLDR(LocaleUtility.getDefaultLocale().toLanguageTag());
+        }
+        return regionMap;
+    }
+
+    private Map<String, String> getRegionsFromDS(String language, ListIterator<DataSourceEnum> msgSourceQueueIter) {
```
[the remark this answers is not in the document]
huihuiw01: Unified the methods' name with 'Region'

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
huihuiw01: As discussed offline, L3 should call ProductService/LocalProductOpt to get supported locales, should not call methods in this class. 
For L2 online mode supportedLanguageList API in Singleton service implement the get supported locales logic, so here L2 needn't call ProductService/LocalProductOpt to get supported locales. 
But for L2 offline mode, need implement the logic of getting supported locales like this.

src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -4,89 +4,107 @@
...
+            logger.debug("Find pattern from cache for locale [{}]!", locale);
+            return patterns;
+        }
+        patterns = getPatternsFromDS(locale, VIPCfg.getInstance().getMsgOriginsQueue().listIterator());
+        if (patterns != null) {
+            logger.debug("Find the pattern for locale [{}].\n", locale);// [datetime] and
+            new PatternCacheService().addPatterns(cacheKey, patterns);
+            logger.debug("Pattern is cached for locale [{}]!\n\n", locale);
+            return patterns;
+        }
```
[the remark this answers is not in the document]
huihuiw01: As explained offline, I think these logic should be here, or no logic in service layer code, but many logic in cache service code, so abort this suggestion. Only move the cache key generation code to PatternCacheService.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -30,66 +31,112 @@ public LocaleService() {
...
+        if (c != null && tmpMap != null) {
+            logger.debug("Find the regions for locale [{}].\n", language);
+            regionMap = JSONUtils.map2SortMap(tmpMap);
+            FormatCacheItem cacheItem = new FormatCacheItem(regionMap);
+            c.put(REGION_PREFIX + language, cacheItem);
+            logger.debug("Regions is cached for locale [{}]!\n\n", language);
+            return regionMap;
+        }
+        if (!LocaleUtility.isDefaultLocale(language)) {
+            logger.info("Can't find regions for locale [{}], look for English regions as fallback!", language);
```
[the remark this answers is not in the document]
huihuiw01: Since the client API exposed to developer is getRegionList, so unified 'Territory' to 'region'.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -80,7 +77,7 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
     
     private List<Locale> getSupportedLocales() {
     	Map<String, String> supportedLanguages = DataSourceEnum.Bundle.createLocaleOpt()
-    			.getLanguages(dto.getLocale());
+    			.getSupportedLanguages(dto.getLocale());
```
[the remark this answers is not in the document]
huihuiw01: I only change the method's name, as discussed offline, here should not call 'getSupportedLanguages' in LocalLocaleOpt, but implement get supported locales in a new class 'LocalProductOpt', and call that.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
```diff
@@ -42,27 +57,98 @@
...
+			return languagesNames;
+		}
+		languagesNames = getLanguagesNamesFromBundle(displayLanguage);
+		if (languagesNames != null) {
+			logger.debug("Find the languages from local bundle for locale [{}].\n", displayLanguage);// [datetime] and
+			new PatternCacheService().addPatterns(cacheKey, languagesNames);
+			logger.debug("Languages is cached for locale [{}]!\n\n", displayLanguage);
+			return languagesNames;
+		}
+		if (!LocaleUtility.isDefaultLocale(displayLanguage)) {
```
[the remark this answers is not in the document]
jessiejuachon: I agree. A fallback logic must be applied only when initial fetch for the requested displayLanguage fails.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
jessiejuachon: @Xiaochao8 , what does the logic here have to do with "product"?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
```diff
@@ -42,27 +57,98 @@
...
+			logger.info("Can't find languages for locale [{}], look for English languages as fallback!", displayLanguage);
+			languagesNames = getLanguagesNames(LocaleUtility.getDefaultLocale().toLanguageTag());
+		}
+		return languagesNames;
+
+	}
+
+	private JSONObject getLanguagesNamesFromBundle(String displayLanguage) {
+		logger.debug("Look for languages from local package bundle for locale [{}]!", displayLanguage);
+		String normalizedLocale = CommonUtil.getCLDRLocale(displayLanguage, localePathMap, localeAliasesMap);
```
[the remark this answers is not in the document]
jessiejuachon: I think it is okay to use CommonUtil.getCLDRLocale in the data (opt) layer. Also, what does it have to do with "pattern".

src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -30,63 +33,90 @@ public JSONObject getPatternsByCategory(String locale, String category) {
 
     public JSONObject getPatterns(String locale) {
```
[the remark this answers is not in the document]
huihuiw01: I tried, but failed.
1. The cache content's type of language and region is Map<String, String>, but pattern's type is Map<String, Object>,
the Map<String, String> in FormatCacheItem can't  be compatible with these two types.
2. The structure of pattern data is nested deeply, and used in many places, it will take a lot of time to change all references.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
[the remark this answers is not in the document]
huihuiw01: I have implemented as above comments.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -30,66 +31,112 @@ public LocaleService() {
...
 
-    public Map<String, String> getDisplayNamesFromCLDR(String language, 
-    		ListIterator<DataSourceEnum> msgSourceQueueIter) {
-    	Map<String, String> dispMap = new HashMap<String, String>(); 	
-    	if (!msgSourceQueueIter.hasNext()) 
-    		return dispMap;
-        
-        logger.trace("look for displayNames from cache");
+    public Map<String, String> getTerritoriesFromCLDR(String language){
+        language = language.replace("_", "-").toLowerCase();
```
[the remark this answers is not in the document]
huihuiw01: Look like no change.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -25,30 +23,30 @@ public ProductService(MessagesDTO dto) {
...
+    public List<String> getSupportedLocales() {
         BaseDTO baseDTO = new BaseDTO();
         baseDTO.setProductID(dto.getProductID());
         baseDTO.setVersion(dto.getVersion());
         ProductBasedOpt dao = new ProductBasedOpt(baseDTO);
-        return dao.getSupportedLocalesFromRemoteVIP();
+        return dao.getSupportedLocales();
     }
 
     public List<Map> getAllComponentTranslation() {
```
Xiaochao8, line 43: Because this method is in ProductService. How about put all translation methods into one service?

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -25,30 +23,30 @@ public ProductService(MessagesDTO dto) {
...
         List<Map> list = new ArrayList<Map>();
         Object[] locales = {};
         Object[] components = {};
```
Xiaochao8, line 46:
```suggestion
    List<String> locales ;
            List<String> components; 
```

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/LocaleOpt.java
```diff
@@ -7,5 +7,6 @@
 import java.util.Map;
 
 public interface LocaleOpt {
-	public Map<String, String> getLanguages(String displayLanguage);
+	public Map<String, String> getSupportedLanguages(String locale);
```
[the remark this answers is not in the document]
huihuiw01: This method is to get supported languages(including language's displayName), corresponding to Singleton service API '/i18n/api/v2/locale/supportedLanguageList' under locale-api, different with get supported locales(only including language tag, no language's displayName). I add 'getSupportedLocales' method in ProductOpt

src/main/java/com/vmware/vipclient/i18n/base/DataSourceEnum.java
```diff
@@ -5,36 +5,65 @@
...
+
 		@Override
-		public LocaleOpt createLocaleOpt() {
-			return new RemoteLocaleOpt();
+		public LocaleOpt createLocaleOpt(BaseDTO dto) {
+			return new RemoteLocaleOpt(dto);
 		}
     };
+    public abstract ProductOpt createProductOpt(BaseDTO dto);
     public abstract MessageOpt createMessageOpt(MessagesDTO dto);
```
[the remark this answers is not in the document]
huihuiw01: I don't make change to this code line, and this is L3 related code, wouldn't like to change them in L2 related PR.
Or code changes may make reviewers confused.

src/main/java/com/vmware/vipclient/i18n/base/DataSourceEnum.java
```diff
@@ -5,36 +5,65 @@
...
-    public abstract LocaleOpt createLocaleOpt();
+    public abstract PatternOpt createPatternOpt();
+    public abstract LocaleOpt createLocaleOpt(BaseDTO dto);
```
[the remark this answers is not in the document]
huihuiw01: Pattern data is format data, not locale data. So they are different, shouldn't be put into one interface.

BTW, on service side, they are are regarded as different resources and provided by different APIs and implemented by different interfaces.

src/main/java/com/vmware/vipclient/i18n/base/DataSourceEnum.java
```diff
@@ -5,36 +5,65 @@
...
     }, 
     VIP {
+        @Override
+        public ProductOpt createProductOpt(BaseDTO dto) {
+            return new ProductBasedOpt(dto);
+        }
+
     	@Override
         public MessageOpt createMessageOpt(MessagesDTO dto) {
             return new ComponentBasedOpt(dto);
```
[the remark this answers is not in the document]
huihuiw01: I don't make change to this code line, and this is L3 related code, wouldn't like to change them in L2 related PR.
Or code changes may make reviewers confused.

src/main/java/com/vmware/vipclient/i18n/base/DataSourceEnum.java
```diff
@@ -5,36 +5,65 @@
...
 public enum DataSourceEnum {
     Bundle {
+        @Override
+        public ProductOpt createProductOpt(BaseDTO dto) {
+            return new LocalProductOpt(dto);
+        }
+
     	@Override
         public MessageOpt createMessageOpt(MessagesDTO dto) {
             return new LocalMessagesOpt(dto);
```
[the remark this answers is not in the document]
huihuiw01: I don't make change to this code line, and this is L3 related code, wouldn't like to change them in L2 related PR.
Or code changes may make reviewers confused.

src/main/java/com/vmware/vipclient/i18n/I18nFactory.java
```diff
@@ -90,7 +90,7 @@ public Message getMessageInstance(Class c, VIPCfg cfg) {
         if (null == cfg) {
             key = c.getCanonicalName();
         } else {
-            key = cfg.getProductName();
+            key = c.getCanonicalName() + "." + cfg.getProductName();
```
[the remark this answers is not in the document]
huihuiw01: Because two different instances(TranslationMessage and LocaleMessage) support shared component, if not add class name, the key for two different instances is the same, one instance maybe covered by the other.

src/main/java/com/vmware/vipclient/i18n/base/instances/LocaleMessage.java
```diff
@@ -71,12 +85,68 @@ public LocaleMessage() {
...
+            return dispMap;
+        }
+        if (!LocaleUtility.isDefaultLocale(locale)) {
+            logger.info("Can't find supported languages for locale [{}], look for English languages as fallback!", locale);
+            Locale fallbackLocale = LocaleUtility.getDefaultLocale();
+            dispMap = localeService.getSupportedDisplayNames(fallbackLocale.toLanguageTag());
+            if (dispMap != null && dispMap.size() > 0) {
+                new FormattingCacheService().addSupportedLanguages(dto, locale, dispMap);
+                logger.debug("Default locale's displayNames is cached for product [{}], version [{}], locale [{}]!\n\n",
+                        dto.getProductID(), dto.getVersion(), locale);
```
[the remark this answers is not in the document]
huihuiw01: This is changed according to Jessiely's suggestion, cache default locale's content as requested locale, or will always try to fetch the requested locale first before using the fallback locale that is already in the cache, which may increase product UI loading time.


Below is Jessiely's original words:

Put this fallback logic in LocaleMessage so that this method here in LocaleService is always for the requested language only.

Once you move the fallback logic into LocaleMessage, update the cache in there in such a way that the key with the requested language is mapped to the FormatCacheItem of the fallback language.
c.put(<cacheKeyOfRequestedLanguage>, new FormatCacheItem(<displayMapofFallbackLanguage>));

Without the above proposed changes, your code will always try to fetch the requested locale first before using the fallback locale that is already in the cache. With the above proposed changes, any succeeding calls for the requested language will immediately return the cached fallback locale until it expires.

Note: You can either put the above logic in LocaleMessage as I propose (just like how it is in TranslationMessage.getMessages), or create another method here in LocaleService to keep the logic in the "service layer".

src/main/java/com/vmware/vipclient/i18n/base/instances/PatternMessage.java
```diff
@@ -26,7 +30,22 @@ public PatternMessage() {
...
+        PatternService ps = new PatternService();
+        patterns = ps.getPatterns(locale.toLanguageTag());
+        if (patterns != null) {
+            return patterns;
+        }
+        if (!LocaleUtility.isDefaultLocale(locale)) {
+            logger.info("Can't find pattern for locale [{}], look for English pattern as fallback!", locale);
+            patterns = ps.getPatterns(LocaleUtility.getDefaultLocale().toLanguageTag());
+            if (patterns != null) {
+                new FormattingCacheService().addPatterns(locale.toLanguageTag(), patterns);
```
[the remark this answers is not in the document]
huihuiw01: Same reason as that for LocaleMessage.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/RemoteProductOpt.java
```diff
@@ -52,7 +52,7 @@ public ProductBasedOpt(BaseDTO dto) {
      * @see com.vmware.vipclient.i18n.messages.dao.ILocaleDao#getSupportedLocales()
      * @deprecated Replaced by {@link com.vmware.vipclient.i18n.messages.api.opt.server.RemoteLocaleOpt#getSupportedLanguages(String)()}
      */
-    @Deprecated public JSONArray getSupportedLocalesFromRemoteVIP() {
+    public List<String> getSupportedLocales() {
         JSONArray msgObject = new JSONArray();
         String responseStr = "";
         Map<String, Object> response = VIPCfg.getInstance().getVipService().getHttpRequester().request(V2URL.getSupportedLocaleListURL(
```
jessiejuachon, line 58: The plan is to deprecate this V2URL on service side because it is but a subset of V2URL.getSupportedLanguageListURL (language tags only instead of tags + display names). So call LocaleService.getDisplayNames(null) in here, just grab the keySet(), and it will achieve the same result.
Note: See Code suggestion in LocaleService.getDisplayNames() first.

src/test/java/com/vmware/vip/i18n/DateFormatUtilOfflineTest.java
```diff
@@ -0,0 +1,302 @@
...
+    final String timeZone = "GMT+8";
+    Date date = new Date(timestamp);
+
+    DateFormatting dateFormatI18n;
+
+    @Before
+    public void init() {
+        VIPCfg gc = VIPCfg.getInstance();
+        try {
+            gc.initialize("vipconfig-offline");
```
jessiejuachon, line 70: This will cause other tests to fail because VIPCFG is a singleton. I had added vipconfig-offline before, but Colin and Xiaochao do not want to use it inside the tests because they say that we should not be able to re-initialize VIPCfg. Hence, that .properties file is not used anymore. Only vipconfig.properties is used, and if I have to modify any property for certain unit tests, I do it inside the test itself. See LocaleTest.testGetSupportedLocalesOfflineBundles for an example.

src/test/java/com/vmware/vip/i18n/MessageFormatOfflineTest.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Locale;
+import java.util.Map;
+
+public class MessageFormatOfflineTest extends BaseTestClass {
+
+    @Before
+    public void init() {
+        VIPCfg gc = VIPCfg.getInstance();
+        try {
+            gc.initialize("vipconfig-offline");
```
jessiejuachon, line 27: Same comment: "This will cause other tests to fail because VIPCFG is a singleton. I had added vipconfig-offline before, but Colin and Xiaochao do not want to use it inside the tests because they say that we should not be able to re-initialize VIPCfg. Hence, that .properties file is not used anymore. Only vipconfig.properties is used, and if I have to modify any property for certain unit tests, I do it inside the test itself. See LocaleTest.testGetSupportedLocalesOfflineBundles for an example. "

src/test/java/com/vmware/vip/i18n/NumberFormatUtilOfflineTest.java
```diff
@@ -0,0 +1,272 @@
...
+
+public class NumberFormatUtilOfflineTest extends BaseTestClass {
+
+    NumberFormatting numberFormatI18n;
+
+    @Before
+    public void init() {
+        VIPCfg gc = VIPCfg.getInstance();
+        try {
+            gc.initialize("vipconfig-offline");
```
jessiejuachon, line 26: Same comment: "This will cause other tests to fail because VIPCFG is a singleton. I had added vipconfig-offline before, but Colin and Xiaochao do not want to use it inside the tests because they say that we should not be able to re-initialize VIPCfg. Hence, that .properties file is not used anymore. Only vipconfig.properties is used, and if I have to modify any property for certain unit tests, I do it inside the test itself. See LocaleTest.testGetSupportedLocalesOfflineBundles for an example. "

src/test/java/com/vmware/vip/i18n/PluralFormatOfflineTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public PluralFormatOfflineTest() {
+        // TODO Auto-generated constructor stub
+    }
+
+    @Before
+    public void init() {
+        VIPCfg gc = VIPCfg.getInstance();
+        try {
+            gc.initialize("vipconfig-offline");
```
jessiejuachon, line 29: Same comment: "This will cause other tests to fail because VIPCFG is a singleton. I had added vipconfig-offline before, but Colin and Xiaochao do not want to use it inside the tests because they say that we should not be able to re-initialize VIPCfg. Hence, that .properties file is not used anymore. Only vipconfig.properties is used, and if I have to modify any property for certain unit tests, I do it inside the test itself. See LocaleTest.testGetSupportedLocalesOfflineBundles for an example. "

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/RemoteProductOpt.java
jessiejuachon, line 58: Just like what I put in the Javadoc comment of this method,  the plan is to deprecate this V2URL endpoint on service side because it is but a subset of V2URL.getSupportedLanguageListURL (language tags only instead of tags + display names). So call RemoteLocaleOpt.getSupportedLanguages(default locale) in here and just grab the keySet().
```suggestion
    public List<String> getSupportedLocales() {
        Iterator<Locale> fallbackLocalesIter = LocaleUtility.getFallbackLocales().iterator();
        return getSupportedLocales(fallbackLocalesIter);
    }
    private List<String> getSupportedLocales(Iterator<Locale> fallbackLocalesIter) {
        if (fallbackLocalesIter != null && fallbackLocalesIter.hasNext()) {
            String locale = fallbackLocalesIter.next().toLanguageTag();
            Map<String, String> supportedLocales = RemoteLocaleOpt.getSupportedLanguages(locale);
            if (supportedLocales == null || supportedLocales.isEmpty()) {
                return getSupportedLocales(fallbackLocalesIter);
            }
            return supportedLocales.keySet();
        }
    }
    
```

src/test/java/com/vmware/vip/i18n/TranslationMessageTest.java
```diff
@@ -178,7 +185,7 @@ public void testGetPatternMessageWithBundle() {
         Assert.assertEquals("\"我的硬盘\"上有345,678个文件。", pluralMessage6);
     }
 
-    @Test
```
Xiaochao8, line 181: It's better to solve the problem instead of commenting it.

src/main/java/com/vmware/vipclient/i18n/base/instances/LocaleMessage.java
```diff
@@ -71,12 +86,37 @@ public LocaleMessage() {
...
+            return null;
+        }
+        LocaleDTO dto = new LocaleDTO();
+        if(cfg != null) {
+            dto.setProductID(cfg.getProductName());
+            dto.setVersion(cfg.getVersion());
+        }else{
+            dto.setProductID(VIPCfg.getInstance().getProductName());
+            dto.setVersion(VIPCfg.getInstance().getVersion());
+        }
```
Xiaochao8, line 110:
```suggestion
  LocaleDTO dto = new LocaleDTO(cfg)
```

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -28,45 +30,18 @@ public ProductService(BaseDTO dto) {
...
     /**
      * Retrieves translated messages of all components of a product in the requested locale (See the dto object).
      *
      * @return translated messages of all components of a product locale specified in the dto object
      */
     public List<Map> getAllComponentTranslation() {
         List<Map> list = new ArrayList<Map>();
-        LocaleDTO localeDTO = new LocaleDTO(dto.getProductID(), dto.getVersion());
-        Map<String, String> locales = new LocaleService(localeDTO).getSupportedLanguages();
+        LocaleDTO localeDTO = new LocaleDTO();
```
Xiaochao8, line 40: Remove this line.

src/main/java/com/vmware/vipclient/i18n/messages/service/FormattingCacheService.java
```diff
@@ -0,0 +1,125 @@
...
+        Map<String, String> o = null;
+        Cache c = VIPCfg.getInstance().getCacheManager()
+                .getCache(VIPCfg.CACHE_L2);
+        if (c != null) {
+        	FormatCacheItem cacheItem = (FormatCacheItem) c.get(key);
+        	if (cacheItem != null)
+        		o = cacheItem.getCachedData();
+        }
+        return o;
+    }
```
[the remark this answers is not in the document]
huihuiw01: This comment is related to cache update, as we discussed offline, cache update will be added in the next PR, so I will resolve this comment in the next PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/FormattingCacheService.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public JSONObject getPatterns(String language, String region) {
+        String cacheKey = getPatternsCacheKey(language, region);
+        return getFormattingPatterns(cacheKey);
+    }
+
+    public Map<String, String> getSupportedLanguages(BaseDTO dto, String locale) {
+        String cacheKey = getSupportedLanguagesCacheKey(dto, locale);
+        return getFormattings(cacheKey);
+    }
```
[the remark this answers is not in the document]
huihuiw01: This comment is related to cache update, as we discussed offline, cache update will be added in the next PR, so I will resolve this comment in the next PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/PatternService.java
```diff
@@ -30,63 +33,90 @@ public JSONObject getPatternsByCategory(String locale, String category) {
...
+                logger.debug("Look for pattern from local bundle for locale [{}]!", locale);
+                String normalizedLocale = CommonUtil.getCLDRLocale(locale, localePathMap, localeAliasesMap);
+                logger.debug("Normalized locale for locale [{}] is [{}]", locale, normalizedLocale);
+                patterns = new LocalPatternOpt().getPatternsByLocale(normalizedLocale);
             }
         }
         return patterns;
     }
 
     private JSONObject getPatternsFromBundle(String language, String region) {
```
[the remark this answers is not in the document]
huihuiw01: Yeah, but this depends on doing locale fallback/normalization first, and currently locale fallback/normalization is based on cldr locale related data. So this means L2 data needed for locale fallback/normalization must be always included in Java client package.

But currently to reduce the size of Java client package, pattern data is put into a separate jar, and for online mode this jar isn't neccessary. If product choose online mode and don't add this jar, change as your suggestion will lead to program broken. 

So I prefer to remove this after we can guarantee data needed for locale fallback/normalization is always included or we define new locale fallback/normalization rule which doesn't relies on cldr data.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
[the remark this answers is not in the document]
huihuiw01: This is locale related code, so your suggestion should be 'do this in LocaleMessage or LocaleService.', right?

The reason of can't resolve it now is the same with below comment:

Don't need this method because client should get the locale from language and region in offline mode.
In general, Getting Pattern by locale is enough.


Currently locale fallback/normalization is based on cldr locale related data. So this means L2 data needed for locale fallback/normalization must be always included in Java client package.

But currently to reduce the size of Java client package, pattern data is put into a separate jar, and for online mode this jar isn't neccessary. If product choose online mode and don't add this jar, change as your suggestion will lead to program broken.

So I prefer to remove this after we can guarantee data needed for locale fallback/normalization is always included or we define new locale fallback/normalization rule which doesn't relies on cldr data.

src/main/java/com/vmware/vipclient/i18n/base/instances/PatternMessage.java
```diff
@@ -4,16 +4,20 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import com.vmware.vipclient.i18n.messages.service.PatternService;
+import java.util.Locale;
 
 /**
  * provide api to get pattern data from remote or locale
  */
 public class PatternMessage implements Message {
```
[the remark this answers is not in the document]
huihuiw01: Yeah, it's required for getPatternMessage, because it's required parameter when call Singleton service API.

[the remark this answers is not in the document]
huihuiw01: resolved:
- moving the fallback and caching logic in to the service layer

not  resolved:
- passing Iterator iter = LocaleUtility.getFallbackLocales().iterator() to the service
I can get the fallback iterator in service layer, no need to get it in message layer and pass it, xiaochao also suggest this.

- using a PatternDTO to hold the locale parameter.
Currently only one parameter 'locale' is needed, and it's better passed as method parameter, so no need to define PatternDTO. As we discussed offline, you are also ok with this.

src/main/java/com/vmware/vipclient/i18n/base/instances/LocaleMessage.java
```diff
@@ -71,12 +85,68 @@ public LocaleMessage() {
...
+        if (!LocaleUtility.isDefaultLocale(locale)) {
+            logger.info("Can't find regions for locale [{}], look for English regions as fallback!", locale);
+            regionMap = localeService.getRegionsFromCLDR(LocaleUtility.getDefaultLocale().toLanguageTag());
+            if (regionMap != null) {
+                new FormattingCacheService().addRegions(locale, regionMap);
+                logger.debug("Default locale's regions is cached for locale [{}]!\n\n", locale);
+            }
+        }
+        return regionMap;
+    }
```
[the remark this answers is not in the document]
huihuiw01: resolved:
Move the locale fallback and caching logic in the service layer

not resolved:
Use LocaleDTO to hold the display locale. Do not pass it as a parameter here.

As we discussed offline,  add a new class LocaleDTO in case parameter increase,  but won’t put locale into it.
Because 'locale' is a frequently changed parameter, put it into DTO will need new service instance multiple times when its value changed, and new instance is time consuming.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -4,92 +4,91 @@
...
-                    + "' data from backend");
-            Map<String, String> tmpMap = new RemoteLocaleOpt()
-				        .getTerritoriesFromCLDR(language);
-            regionMap = JSONUtils.map2SortMap(tmpMap);
-            respMap.put(language, regionMap);
-            if (c != null) {
-            	FormatCacheItem cacheItem = new FormatCacheItem(regionMap);
-                c.put(REGION_PREFIX + language, cacheItem);
-            }
+    public Map<String, String> getRegionsFromCLDR(String locale){
```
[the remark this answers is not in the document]
huihuiw01: - Remove locale parameter, make it a property of LocaleDTO.
As we discussed offline, add a new class LocaleDTO in case parameter increase, but won’t put locale into it.
Because 'locale' is a frequently changed parameter, put it into DTO will need new service instance multiple times when its value changed, and new instance is time consuming.

- Use fallbackLocalesIter as parameter and recursively call the method to apply locale fallback logic.
I can get the fallback iterator in service layer, no need to get it in message layer and pass it, xiaochao also suggest this.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -4,92 +4,91 @@
...
+        //cacheItem = new FormatCacheItem();
+        dispMap = getSupportedLanguagesFromDS(locale, VIPCfg.getInstance().getMsgOriginsQueue().listIterator());
+        if (dispMap != null && dispMap.size() > 0) {
+            logger.debug("Find the displayNames for product [{}], version [{}], locale [{}].\n", productName, version, locale);
+            formattingCacheService.addSupportedLanguages(dto, locale, dispMap);
+            logger.debug("DisplayNames is cached for product [{}], version [{}], locale [{}]!\n\n", productName, version, locale);
+            return dispMap;
         }
         return dispMap;
     }
```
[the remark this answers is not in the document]
huihuiw01: As we discussed offline, will add cache update in the next PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -4,179 +4,125 @@
...
-        
+    private Map<String, String> getSupportedLanguagesFromDS(String locale, ListIterator<DataSourceEnum> msgSourceQueueIter) {
+        Map<String, String> dispMap = new HashMap<String, String>();
+        if (!msgSourceQueueIter.hasNext())
+            return dispMap;
+        DataSourceEnum dataSource = (DataSourceEnum) msgSourceQueueIter.next();
+        dispMap = dataSource.createLocaleOpt(dto).getSupportedLanguages(locale);
         if (dispMap == null || dispMap.isEmpty()) {
-        	return getDisplayNamesFromCLDR(language, msgSourceQueueIter);
+            dispMap = getSupportedLanguagesFromDS(locale, msgSourceQueueIter);
```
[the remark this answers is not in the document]
huihuiw01: Actually I resolve this in cache update commit, since I plan to put cache update in a new PR, so I reset that commit, after reset the change reset, too. So I resolve this in this PR again.

src/test/java/com/vmware/vip/i18n/DateFormatUtilOfflineTest.java
huihuiw01: As we discussed offline, unit tests is for testing different scenarios, not for mock real scenarios,  re-initialize is acceptable. And use  properties is more convenient, or we need do setMsgOriginsQueue in many places to re-initialize data source.

So I clear data source queue before each initiallization by calling clearDataSource method in the constructor of parent class of all test classes 'BaseTestClass', and problem is resolved.

src/test/java/com/vmware/vip/i18n/MessageFormatOfflineTest.java
huihuiw01: Same resolution as above.

src/test/java/com/vmware/vip/i18n/NumberFormatUtilOfflineTest.java
huihuiw01: Same resolution as above.

src/test/java/com/vmware/vip/i18n/PluralFormatOfflineTest.java
huihuiw01: Same resolution as above.

src/main/java/com/vmware/vipclient/i18n/base/instances/LocaleMessage.java
```diff
@@ -33,8 +43,15 @@ public LocaleMessage() {
      * @return map the key is the supported language tag, value is all region data mapped by
      *         language tag
      */
-    public Map<String, Map<String, String>> getRegionList(List<String> supportedLanguageList) {
-        return new LocaleService().getTerritoriesFromCLDR(supportedLanguageList);
+    public Map<String, Map<String, String>> getRegionList(List<String> localeList) {
+        Map<String, Map<String, String>> respMap = new HashMap<String, Map<String, String>>();
+        for (String locale : localeList) {
+            if(locale != null && !locale.isEmpty()) {
+                Map<String, String> regionMap = getRegionsFromCLDR(locale);
```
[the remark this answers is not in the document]
huihuiw01: As we discussed offline, add a new class LocaleDTO in case parameter increase, but won’t put locale into it.
Because 'locale' is a frequently changed parameter, put it into DTO will need new service instance multiple times when its value changed, and new instance is time consuming.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -4,92 +4,91 @@
...
+        DataSourceEnum dataSource = (DataSourceEnum) msgSourceQueueIter.next();
+        regions = dataSource.createLocaleOpt(dto).getRegions(locale);
+        if (regions == null || regions.isEmpty()) {
+            regions = getRegionsFromDS(locale, msgSourceQueueIter);
         }
-        
-        if (dispMap == null || dispMap.isEmpty()) {
-        	return getDisplayNamesFromCLDR(language, msgSourceQueueIter);
+        return regions;
+    }
```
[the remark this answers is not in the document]
huihuiw01: As we discussed offline, add a new class LocaleDTO in case parameter increase, but won’t put locale into it.
Because 'locale' is a frequently changed parameter, put it into DTO will need new service instance multiple times when its value changed, and new instance is time consuming.

src/main/java/com/vmware/vipclient/i18n/base/instances/LocaleMessage.java
```diff
@@ -56,12 +71,12 @@ public LocaleMessage() {
      * @return a list contains the supported language tags
      */
     public List<String> getSupportedLanguageTagList() {
-        Map<String, String> languageTagMap = new LocaleService()
-                .getDisplayNamesFromCLDR(java.util.Locale.ENGLISH.toLanguageTag(), 
-                		VIPCfg.getInstance().getMsgOriginsQueue().listIterator());
-        Collection<String> keyCollection = languageTagMap.keySet();
-        List<String> languageTagList = new ArrayList<String>(keyCollection);
-        return languageTagList;
+        Map<String, String> languageTagMap = getDisplayNamesFromCLDR(LocaleUtility.getDefaultLocale().toLanguageTag());
```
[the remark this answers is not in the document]
huihuiw01: Could you give me an example in what kind of situation there maybe more than one default locale?

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -106,4 +81,28 @@ public JSONArray getSupportedLocalesFromRemoteVIP() {
...
+    public List<String> getSupportedLocales(Iterator<DataSourceEnum> msgSourceQueueIter) {
+        if (!msgSourceQueueIter.hasNext()) { return null; }
+
+        DataSourceEnum dataSource = msgSourceQueueIter.next();
+        ProductOpt opt = dataSource.createProductOpt(dto);
+        List<String> locales = opt.getSupportedLocales();
+        // If failed to get components from the data source
+        if (locales.isEmpty()) {
+            // Try the next dataSource in the queue
+            if (msgSourceQueueIter.hasNext()) {
```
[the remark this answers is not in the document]
huihuiw01: Write the same with you, just copy it from ComponentService.

src/main/java/com/vmware/vipclient/i18n/base/instances/LocaleMessage.java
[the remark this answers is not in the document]
Xiaochao8: It looks this method can be removed because nobody uses it...

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/RemoteProductOpt.java
huihuiw01: I prefer to deprecate this V2URL endpoint after V3 API is available.
Looks like V3 API url is different with V2, after V3 is available, this need to be changed again.

src/main/java/com/vmware/vipclient/i18n/base/instances/LocaleMessage.java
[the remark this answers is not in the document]
huihuiw01: Yeah, maybe nobody use it, but since it's in a class exposed to user, so keep it.
But we may needn't spend much time or discussion on it.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -4,179 +4,125 @@
...
+        }
+        if (!LocaleUtility.isDefaultLocale(locale)) {
+            logger.info("Can't find supported languages for locale [{}], look for default locale's languages as fallback!", locale);
+            Locale fallbackLocale = LocaleUtility.getDefaultLocale();
+            dispMap = getSupportedDisplayNamesByLocale(fallbackLocale.toLanguageTag());
+            if (dispMap != null && dispMap.size() > 0) {
+                new FormattingCacheService().addSupportedLanguages(dto, locale, dispMap);
+                logger.debug("Default locale's displayNames is cached for product [{}], version [{}], locale [{}]!\n\n",
+                        dto.getProductID(), dto.getVersion(), locale);
             }
```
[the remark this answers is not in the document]
huihuiw01: Will do in the next PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -4,179 +4,125 @@
...
+        if (!LocaleUtility.isDefaultLocale(locale)) {
+            logger.info("Can't find regions for locale [{}], look for default locale's regions as fallback!", locale);
+            regionMap = getRegionsByLocale(LocaleUtility.getDefaultLocale().toLanguageTag());
+            if (regionMap != null) {
+                new FormattingCacheService().addRegions(locale, regionMap);
+                logger.debug("Default locale's regions is cached for locale [{}]!\n\n", locale);
+            }
+        }
+        return regionMap;
     }
```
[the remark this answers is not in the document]
huihuiw01: Will do in the next PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -58,23 +56,17 @@ public ProductService(BaseDTO dto) {
...
-            } else {
+    public List<String> getComponents(){
+        List<String> components = null;
+        Iterator<DataSourceEnum> msgSourceQueueIter = VIPCfg.getInstance().getMsgOriginsQueue().iterator();
+        while((components == null || components.isEmpty()) && msgSourceQueueIter.hasNext()){
+            DataSourceEnum dataSource = msgSourceQueueIter.next();
+            ProductOpt opt = dataSource.createProductOpt(dto);
+            components = opt.getComponents();
+            // If failed to get components from the data source, log the error.
+            if (components.isEmpty()) {
```
Xiaochao8, line 69:
```suggestion
            if (components == null || components.isEmpty()) {
```

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
```diff
@@ -96,37 +95,33 @@ private JSONObject getLanguagesNamesFromBundle(String locale) {
...
 		String normalizedLocale = CommonUtil.getCLDRLocale(locale, localePathMap, localeAliasesMap);
 		logger.info("Normalized locale for locale [{}] is [{}]", locale, normalizedLocale);
 		if(normalizedLocale == null || normalizedLocale.isEmpty())
 			return null;
-		String regionsJsonStr = PatternUtil.getRegionFromLib(normalizedLocale);
-		try {
+        try {
+            JSONObject regionsData = null;
+		    String regionsJsonStr = PatternUtil.getRegionFromLib(normalizedLocale);
 			regionsData = (JSONObject) new JSONParser().parse(regionsJsonStr);
```
Xiaochao8:
```suggestion
		    String regionsJsonStr = PatternUtil.getRegionFromLib(normalizedLocale);
			JSONObject regionsData = (JSONObject) new JSONParser().parse(regionsJsonStr);
```

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalPatternOpt.java
```diff
@@ -42,14 +41,15 @@ public JSONObject getPatterns(String language, String region) {
     private JSONObject getPatternsByLocale(String normalizedLocale) {
         if(normalizedLocale == null || normalizedLocale.isEmpty())
             return null;
-        String patternStr = PatternUtil.getPatternFromLib(normalizedLocale, null);
-        Map<String, Object> patterns = null;
         try {
+            Map<String, Object> patterns = null;
+            String patternStr = PatternUtil.getPatternFromLib(normalizedLocale, null);
             patterns = (Map<String, Object>) new JSONParser().parse(patternStr);
```
Xiaochao8:
```suggestion
            String patternStr = PatternUtil.getPatternFromLib(normalizedLocale, null);
            Map<String, Object> patterns = (Map<String, Object>) new JSONParser().parse(patternStr);
```

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -116,11 +124,14 @@ public LocaleService(LocaleDTO dto) {
 
     private Map<String, String> getSupportedLanguagesFromDS(String locale, ListIterator<DataSourceEnum> msgSourceQueueIter) {
         Map<String, String> dispMap = new HashMap<String, String>();
-        if (!msgSourceQueueIter.hasNext())
+        if (!msgSourceQueueIter.hasNext()) {
+            logger.error(FormatUtils.format(ConstantsMsg.GET_LANGUAGES_FAILED_ALL));
```
Xiaochao8, line 128:
```suggestion
            logger.error(ConstantsMsg.GET_LANGUAGES_FAILED_ALL);
```

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -116,11 +124,14 @@ public LocaleService(LocaleDTO dto) {
...
             return dispMap;
+        }
         DataSourceEnum dataSource = (DataSourceEnum) msgSourceQueueIter.next();
         dispMap = dataSource.createLocaleOpt(dto).getSupportedLanguages(locale);
         if (dispMap == null || dispMap.isEmpty()) {
+            logger.debug(FormatUtils.format(ConstantsMsg.GET_LANGUAGES_FAILED, dataSource.toString()));
```
Xiaochao8, line 134: Suggest log in this way. FromatUtils is heavy and unnecessary.

```suggestion
            logger.debug(ConstantsMsg.GET_LANGUAGES_FAILED, dataSource);
```

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -37,10 +36,9 @@ public ProductService(BaseDTO dto) {
...
-        LocaleDTO localeDTO = new LocaleDTO();
-        List<String> locales = this.getSupportedLocales(VIPCfg.getInstance().getMsgOriginsQueue().iterator());
-        List<String> components = this.getComponents(VIPCfg.getInstance().getMsgOriginsQueue().iterator());
-        if (locales != null) {
+        List<String> locales = this.getSupportedLocales();
+        List<String> components = this.getComponents();
+        if (locales != null && components != null) {
             for (String languageTag : locales) {
                 for (Object component : components) {
                     MessagesDTO msgDTO = new MessagesDTO(((String) component).trim(), LocaleUtility.fmtToMappedLocale(Locale.forLanguageTag(languageTag)).toString().trim(),
```
Xiaochao8, line 44: It's sure that component and locale are normalized. Don't need to do further procession.
```suggestion
                    MessagesDTO msgDTO = new MessagesDTO(component, languageTag, dto.getProductID(), dto.getVersion());
```

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -84,23 +76,18 @@ public ProductService(BaseDTO dto) {
...
-                logger.error(FormatUtils.format(ConstantsMsg.GET_COMPONENTS_FAILED, dataSource.toString()));
+    public List<String> getSupportedLocales(){
+        List<String> locales = null;
+        Iterator<DataSourceEnum> msgSourceQueueIter = VIPCfg.getInstance().getMsgOriginsQueue().iterator();
+        while((locales == null || locales.isEmpty()) && msgSourceQueueIter.hasNext()){
+            DataSourceEnum dataSource = msgSourceQueueIter.next();
+            ProductOpt opt = dataSource.createProductOpt(dto);
+            locales = opt.getSupportedLocales();
+            // If failed to get locales from the data source, log the error.
+            if (locales.isEmpty()) {
```
Xiaochao8, line 89:
```suggestion
            if (locales==null || locales.isEmpty()) {
```

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -84,23 +76,18 @@ public ProductService(BaseDTO dto) {
...
+                logger.error(FormatUtils.format(ConstantsMsg.GET_LOCALES_FAILED, dataSource.toString()));
```
Xiaochao8, line 90: FormatUtils.format is heavy and unnecessary.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
huihuiw01: Actually ‘components’ won't be null as opt.getComponents() method new ArrayList and won't return null.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
huihuiw01: Actually ‘locales’ won't be null as opt.getSupportedLocales() method new ArrayList and won't return null.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
huihuiw01: Don't make this change as this is L3 code and not sure whether there will be error or not if don't do such process.

src/main/java/com/vmware/vipclient/i18n/messages/service/LocaleService.java
```diff
@@ -4,179 +4,136 @@
...
+            return dispMap;
+        }
+        if (locale != null && !locale.isEmpty() && !LocaleUtility.isDefaultLocale(locale)) {
+            logger.info("Can't find supported languages for locale [{}], look for default locale's languages as fallback!", locale);
+            Locale fallbackLocale = LocaleUtility.getDefaultLocale();
+            dispMap = getSupportedDisplayNamesByLocale(fallbackLocale.toLanguageTag());
+            if (dispMap != null && dispMap.size() > 0) {
+                new FormattingCacheService().addSupportedLanguages(dto, locale, dispMap);
+                logger.debug("Default locale's displayNames is cached for product [{}], version [{}], locale [{}]!\n\n",
+                        dto.getProductID(), dto.getVersion(), locale);
```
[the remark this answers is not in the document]
huihuiw01: - Use of fallbackLocalesIter
resolved

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+        } catch (Exception e) {
+            logger.debug(e.getMessage());
+        }
+        return components;
+    }
+
+    private void getSupportedLocales(Path path, List<String> supportedLocales) throws IOException {
+        try (Stream<Path> listOfFiles = Files.walk(path).filter(p -> Files.isRegularFile(p))) {
```
jessiejuachon, line 87: @huihuiw01 , you said you had to change this back to isRegulsrFile for the smoke test to pass. I see isRegularFile in current LocalLocaleOpt.java though:
https://github.com/vmware/singleton/blob/g11n-java-client/src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
so how did the erroneous code get to your PR?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
```diff
@@ -0,0 +1,113 @@
...
+            listOfFiles.map(file -> {
+                return file.getFileName().toString();
+            }).forEach(component -> {
+                if(component != null && !component.isEmpty()) {
+                    components.add(component);
+                }
+            });
+        }
+    }
+}
```
jessiejuachon, line 113:
```suggestion
    private void getComponents(Path path, List<String> components) throws IOException {
        try (Stream<Path> listOfFiles = Files.list(path).filter(p -> !Files.isRegularFile(p))) {
            listOfFiles.forEach(s -> components.add(s.getFileName().toString()));
        }
    }
```

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
```diff
@@ -0,0 +1,113 @@
...
+                    return fileName.substring(BUNDLE_PREFIX.length(), fileName.indexOf('.'));
+                }
+                return "";
+            }).forEach(language -> {
+                if(language != null && !language.isEmpty()) {
+                    supportedLocales.add(language);
+                }
+            });
+        }
+    }
```
jessiejuachon, line 100:
```suggestion
    private void getSupportedLocales(Path path, List<String> supportedLocales) throws IOException {
        try (Stream<Path> listOfFiles = Files.walk(path).filter(p -> Files.isRegularFile(p) && p.getFileName().startsWith(BUNDLE_PREFIX) && p.getFileName().endsWith(BUNDLE_SUFFIX))) {
            listOfFiles.map(file -> {
                return file.getFileName().toString().substring(BUNDLE_PREFIX.length(), fileName.indexOf('.'));
            }).forEach(language -> {
                supportedLocales.add(language);
            });
        }
    }
```

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
```diff
@@ -4,65 +4,122 @@
...
+		if (!LocaleUtility.isDefaultLocale(locale)) {
+			logger.info("Can't find languages' names for locale [{}], look for English regions as fallback!", locale);
+			languagesNames = getLanguagesNamesFromCLDR(LocaleUtility.getDefaultLocale().toLanguageTag());
+			if (languagesNames != null) {
+				new FormattingCacheService().addLanguagesNames(locale, languagesNames);
+				logger.debug("Default locale's languages' names is cached for locale [{}]!\n\n", locale);
+			}
+		}
+		return languagesNames;
+	}*/
```
jessiejuachon, line 71: Can this be removed since it is commented out?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
huihuiw01: These two conditions are for filtering file name, not for filter path,
fileName.getFileName().startsWith(BUNDLE_PREFIX) && fileName.getFileName().endsWith(BUNDLE_SUFFIX)

or locale list will be empty because no path meet these conditions.

Xiaochao8 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: I made a mistake of not adding .toString() to p.getFileName(). I have updated the code suggestion.

Files.walk will iterate over ALL paths -- all folders AND files inside the root path and  the root path itself. So yes, it will apply to files as well.  The .filter part will check if each item is a file (not a folder) that starts with BUNDLE_PREFIX and ends with BUNDLE_SUFFIX. If one of these conditions fail, the item will not be included in listOfFiles. This way, you do not need the if statement. 

You also do not need to return an empty String so that you can remove `if(language != null && !language.isEmpty())`

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private void getComponents(Path path, List<String> components) throws IOException {
+        try (Stream<Path> listOfFiles = Files.walk(path).filter(p -> !Files.isRegularFile(p))) {
```
jessiejuachon, line 103: Use Files.list here because Files.walk iterates recursively which we do not need here.
```suggestion
        try (Stream<Path> listOfFiles = Files.list(path).filter(p -> !Files.isRegularFile(p))) {
```

build.gradle
```diff
@@ -1,3 +1,7 @@
+/*
+ * Copyright 2019 VMware, Inc.
```
Xiaochao8, line 2: Make sure new copyright is correct

PR #650, opened 2020-07-07
fix error of GetLocaleList

jin-jasper:


singletonclient/SingletonClient/Implementation/SingletonConfig.cs
```diff
@@ -246,8 +246,11 @@ public IConfigItem GetComponentAttribute(string component, string key)
...
-            IConfigItem localesItem = GetComponentAttribute(component, ConfigConst.KeyLocales);
-            if (localesItem == null)
+            IConfigItem componentItem = GetComponentArrayItem(component);
+            IConfigItem localesItem = (componentItem == null) ?
+                null : componentItem.GetMapItem(ConfigConst.KeyLocales);
+
+            if (componentItem != null && localesItem == null)
             {
                 localesItem = GetItem(ConfigConst.KeyLocales);
             }
```
Xiaochao8, line 256: Could you explain the logic? in the standup meeting?

PR #653, opened 2020-07-08
Fix CI failure of the last commit of Named args

Xiaochao8:


src/main/java/com/vmware/vipclient/i18n/util/FormatUtils.java
```diff
@@ -35,15 +29,11 @@ public static String format(String pattern, Object... arguments) {
         return java.text.MessageFormat.format(escaped, arguments);
     }
 
-    @Deprecated
-    public static String formatMsg(String pattern, Locale locale, Object... arguments) {
-    	if (pattern != null && !pattern.isEmpty() && arguments != null && arguments.length > 0) {
```
huihuiw01, line 40: Will the program throw exception without this judgement?
Why remove this judgement?

Xiaochao8: This code doesn't exist before. It's import in last commit wrong. So here just remove it to revert to the original.

Issue #651, opened 2020-07-08
[BUG] [Service] missing settings.gradle under tools\tool-cldr-extractor


nannany00:
**Describe the bug**
When try to compile a formatting pattern jar package under *tools\tool-cldr-extractor*, it always fails, due to miss **settings.gradle** file.

**To Reproduce**
1. Git clone the latest Singleton Service code
2. Go to *g11n-ws\tools\tool-cldr-extractor*
3. Run `gradle build`
```
Project directory '.\singleton\g11n-ws\tools\tool-cldr-extractor' is not part of the build defined by settings file 'D:\git\github\singleton\g11n-ws\settings.gradle'. If this is an unrelated build, it must have its own settings file.
```

**Expected behavior**
Everything runs well, and the formmating pattern jar package is generated well.

**Screenshots**
NA

**Desktop (please complete the following information):**
NA

**Additional context**
NA

PR #654, opened 2020-07-09
[BUG] [Service] missing settings.gradle under tools\tool-cldr-extractor #651

chaoq0318:
fix issue: https://github.com/vmware/singleton/issues/651


g11n-ws/tools/tool-cldr-extractor/src/main/java/com/vmware/i18n/pattern/service/impl/PatternServiceImpl.java
```diff
@@ -131,16 +131,22 @@ private String parseRegion(String[] arr){
 		Map<String, Object> suppleMap = new HashMap<>();
 		IPatternDao dao = new PatternDaoImpl();
 		String[] cateList = categories.split(",");
+
+		String jsonPath = CLDRConstants.RESOURCES_PATH;
```
Xiaochao8, line 135:
```suggestion
		String resourcePath = CLDRConstants.RESOURCES_PATH;
```

chaoq0318: fixed

PR #682, opened 2020-07-21
Add methods in TranslationMessage to support named arguments

Xiaochao8:


src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -319,11 +357,50 @@ public String getString2(final String component,
...
+        String message = "";
+        String source;
+        try {
+            ResourceBundle rb = ResourceBundle.getBundle(bundle, LocaleUtility.defaultLocale);
+            source = rb.getString(key);
+        } catch (Exception e) {
+            logger.error(e.getMessage());
             source = key;
         }
         // get translation from VIP service
```
linr211, line 402: Can we have another private method for these codes and both getString2 can share it?

Issue #692, opened 2020-07-27
[BUG][Go Client] Offline fetch is triggered when cache item is expired and online bundle fetch failed.


dleng-vmware:
**Describe the bug**
Offline fetch is triggered when cache item is expired and online bundle fetch failed.

**To Reproduce**
1. Set config as below:
"online_service_url": "http://localhost:8091",
"offline_resources_base_url": "../testdata/localBundles",
"default_locale": "zh-Hans"
2. The component "componentlocale" exist in online and offline:
ommsg1, _ := translation.GetComponentMessages(name, version, "fr", "componentlocale")
3. Get one key's value from step2 returned result.
value, _ := commsg.Get("language")
4. Close the Singleton service side and wait the cache expired
5. Repeat the request in step 2.
6. Repeat the request in step 2.
7. Repeat step 3.
Get the offline translation rather than the online translation

**Expected behavior**
The step 7 should get online translation in cache, offline fetch should not be triggered, Overwrite the cache only if online bundle fetch is successful.

dleng-vmware:
After verification, the bug has been fixed

PR #696, opened 2020-08-03
Fix bug https://github.com/vmware/singleton/issues/692

Xiaochao8:
Fix bug https://github.com/vmware/singleton/issues/692


common.go
```diff
@@ -5,7 +5,12 @@
...
-import "fmt"
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+)
 
 // ComponentMsgs The interface of a component's messages
 type ComponentMsgs interface {
```
tigershi, line 16: why define fields in interface？

Issue #646, opened 2020-07-03
[BUG] [Java Client]Get translation with multiple components interface hasn't been marked as deprecated.


lyiyu66:
**Describe the bug**
Get translation with multiple components interface hasn't been marked as deprecated.
 getStrings() Map<Locale, Map<String, Map<String, String>>>

As we plan to deprecate above interface according to https://github.com/vmware/singleton/issues/554#issuecomment-631130215, please help to mark it as deprecated in code.

**Expected behavior**
The API of get translation with multiple components are marked as deprecated in code.

PR #699, opened 2020-08-03
G11n java client - Fix for expiration-related issues (issues 604, 664, and 662)

jessiejuachon:


src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -125,12 +125,19 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+					1. A previously stored etag, if any, can be used for the next HTTP request.
+					2. CacheItem properties such as etag, timestamp and maxAgeMillis can be refreshed
+						with new values from the next HTTP response.
+				*/
+				String cacheItemLocale = cacheItem.getLocale();
+    			if (cacheItemLocale.equals(this.dto.getLocale())) {
+					fetchMessages(cacheItem, VIPCfg.getInstance().getMsgOriginsQueue().listIterator());
+				} else {  //If cacheItem's locale is not the requested locale, it means it is for a fallback locale. Hence, refresh the appropriate fallback locale's cache
+					MessagesDTO fallbackLocaleDTO = new MessagesDTO(dto.getComponent(), cacheItemLocale, dto.getProductID(), dto.getVersion());
+					new ComponentService(fallbackLocaleDTO).fetchMessages(cacheItem, VIPCfg.getInstance().getMsgOriginsQueue().listIterator());
```
Xiaochao8, line 138: For some reasons, the requested locale failed last time. We still need to try the requested locale first, then fallback to default locale if failure.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -67,9 +67,7 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
 				path = Paths.get(uri);
 				messages = JSONBundleUtil.getMessages(path);
 			}
-				
-	    	cacheItem.addCachedData(messages);
-	    	cacheItem.setTimestamp(System.currentTimeMillis());
+			cacheItem.setCacheItem(dto.getLocale(), messages, null, System.currentTimeMillis(), null);
```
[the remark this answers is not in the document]
jessiejuachon: Data from local must expire. Use case: If data from local was fetched as a fallback for when Singleton service is temporarily down. In this case, when the cached data from local expires, it should try to fetch from the service again.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
```diff
@@ -52,33 +50,32 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
...
+				        	cacheItem.setCacheItem(this.dto.getLocale(), messages, etag, timestamp, maxAgeMillis);
 				        }
         			}
 	        	} catch (Exception e) {
 	        		logger.error("Failed to get messages");
 	        	}
-	        }
-
+	        } else {
+                cacheItem.setCacheItem(this.dto.getLocale(), etag, timestamp, maxAgeMillis);
```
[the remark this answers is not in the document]
jessiejuachon: The else is not for when it fails. It is for when status is 304 (HTTP_NOT_MODIFIED). See lines 51-52.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentsService.java
```diff
@@ -96,7 +96,7 @@ public ComponentsService(final VIPCfg config) {
             final MessagesDTO dto = new MessagesDTO();
             dto.setComponent(comp);
             dto.setLocale(locale);
-            new CacheService(dto).addCacheOfComponent(new MessageCacheItem(messages, cacheItem.getEtag(), 
+            new CacheService(dto).addCacheOfComponent(new MessageCacheItem(locale, messages, cacheItem.getEtag(),
```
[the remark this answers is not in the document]
jessiejuachon: Right! Good catch. I'll set the etag to null for this one for now. 
https://github.com/vmware/singleton/issues/646

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: I will update line 103 with this: `if (cacheItem.isExpired() || !cacheItem.getLocale().equals(this.dto.getLocale()))`

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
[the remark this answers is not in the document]
Xiaochao8: It should be set the time in other place because here is Local Opt, it doesn't need to consider the time.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
Xiaochao8: I'm afraid it can't. If one locale gets the default locale's messages, it will always get the default locale's messages. It's wrong.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -99,7 +99,8 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
     	MessageCacheItem cacheItem = null;
     	if (cacheService.isContainComponent()) { // Item is in cache
     		cacheItem = cacheService.getCacheOfComponent();
-    		if (cacheItem.isExpired()) { // cacheItem has expired
+			// If the cacheItem is either expired or for a fallback locale
+    		if (cacheItem.isExpired() || !cacheItem.getLocale().equals(this.dto.getLocale())) {
```
huihuiw01, line 103: With this condition !cacheItem.getLocale().equals(this.dto.getLocale()),
cache default locale's content for requested locale may not take effect, because they never be equal.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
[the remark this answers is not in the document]
jessiejuachon: So where do you recommend to set the expire time?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: Right! My mistake. See the changes please

Xiaochao8: I mean that the requested locale fails not because locale isn't supported but other reasons such as network error. So next time to update cache, still need to get the requested locale first.

jessiejuachon: Yes, that is what is happening as I explained to Huihui. Let's connect over Slack

jessiejuachon: I have also updated the code to make it easier to understand, but same logic.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -100,44 +106,78 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+    private MessageCacheItem createCacheItem(Iterator<Locale> fallbackLocalesIter) {
+		CacheService cacheService = new CacheService(dto);
+		// Create a new cacheItem object to be stored in cache
+		MessageCacheItem cacheItem = new MessageCacheItem();
+		refreshCacheItem(cacheItem, VIPCfg.getInstance().getMsgOriginsQueue().iterator());
+
+		if (!cacheItem.getCachedData().isEmpty()) {
+			cacheService.addCacheOfComponent(cacheItem);
+		} else if (!dto.getLocale().equals(ConstantsKeys.SOURCE) && fallbackLocalesIter!=null && fallbackLocalesIter.hasNext()) {
+			// If failed to fetch message for the requetsed DTO, use MessageCacheItem of the next fallback locale.
```
Xiaochao8, line 137: Need to judge if current fallback locale is same as the locale in DTO. If same, should skip it.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -100,44 +106,78 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
-    
-	private void populateCacheTask(MessageCacheItem cacheItem) {
+
+	/**
+	 * Creates a new MessageCacheItem for the DTO and stores it in cache.
+	 *
+	 * @param fallbackLocalesIter The fallback locale queue to use in case of failure.
+	 *
+	 */
+    private MessageCacheItem createCacheItem(Iterator<Locale> fallbackLocalesIter) {
```
Xiaochao8, line 128: It seems here has a recursive call: getMessages(Iterator<Locale> fallbackLocalesIter)  -> createCacheItem(Iterator<Locale> fallbackLocalesIter)  -> new ComponentService(fallbackLocaleDTO).getMessages(fallbackLocalesIter)

In my opinion, the hierarchical relation between methods are not clear.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -100,44 +106,78 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
-    			// If failed to fetch message, use MessageCacheItem of the next fallback locale.
-				MessagesDTO fallbackLocaleDTO = new MessagesDTO(dto.getComponent(), fallbackLocalesIter.next().toLanguageTag(), dto.getProductID(), dto.getVersion());
-				cacheItem = new ComponentService(fallbackLocaleDTO).getMessages(fallbackLocalesIter);
-				if (!cacheItem.getCachedData().isEmpty()) {
-					cacheService.addCacheOfComponent(cacheItem);
-				}
+			// If the cacheItem is for a fallback locale, create and store cacheItem for the requested locale in a separate thread.
+    		if (!LocaleUtility.isSameLocale(cacheItem.getLocale(), this.dto.getLocale())) {
+				this.createCacheItemTask(fallbackLocalesIter);
 			}
```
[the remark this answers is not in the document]
jessiejuachon: We actually want to always try to fetch every time in a separate thread, not just when the cacheItem has expired. This is because if the cacheItem is for a fallback locale, many other calls for non-supported locales can trigger a refresh  of the fallback locale cacheItem. In that case, it may never enter this if condition if we put it inside the above if block,

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: What is not clear about it? Could you be more specific?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: If you follow the recursive logic, the locale in the dto will be the previous locale in the fallback locale queue.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -100,44 +106,78 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+
+		if (!cacheItem.getCachedData().isEmpty()) {
+			cacheService.addCacheOfComponent(cacheItem);
+		} else if (!dto.getLocale().equals(ConstantsKeys.SOURCE) && fallbackLocalesIter!=null && fallbackLocalesIter.hasNext()) {
+			// If failed to fetch message for the requetsed DTO, use MessageCacheItem of the next fallback locale.
+			MessagesDTO fallbackLocaleDTO = new MessagesDTO(dto.getComponent(), fallbackLocalesIter.next().toLanguageTag(), dto.getProductID(), dto.getVersion());
+			cacheItem = new ComponentService(fallbackLocaleDTO).getMessages(fallbackLocalesIter);
+			if (!cacheItem.getCachedData().isEmpty()) {
+				cacheService.addCacheOfComponent(cacheItem);
+			}
```
huihuiw01, line 142: Here only use the first fallback locale in the fallback locale chain, doesn't go through the whole fallback locale chain.

jessiejuachon: It actually does go through the whole cahin/queue. In line 138, you see `fallbackLocalesIter.next()`. By calling .next(), the current pointer of fallbackLocalesIter moves to the next item in the queue. When you pass fallbackLocalesIter to ComponentService.getMessages() in line139, the iterator holds the correct place (current pointer) in the queue.

PR #713, opened 2020-08-12
Getting from service when source is unavailable

Xiaochao8:


src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -167,7 +221,7 @@ private String getStringWithArgs(final Locale locale,
                     translation = FormatUtils.formatMsg(translation,
                             LocaleUtility.defaultLocale, (Map<String, Object>) args);
                 } else {
-                    translation = FormatUtils.formatMsg(translation, locale, (Map) args);
+                    translation = FormatUtils.formatMsg(translation, locale, (Map<String, Object>) args);
                 }
             }
```
linr211, line 226: 1. How about move these codes to a new private method: getStringWithSource
2. Please try to combine common codes of getStringWithSource and getStringWithoutSource, i think there are some duplicated codes

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -175,6 +189,24 @@ private String getStringWithArgs(final Locale locale,
...
     }
 
+    private String getStringWithoutSource(MessagesDTO dto, final Object args) {
+        Locale locale = Locale.forLanguageTag(dto.getLocale());
+        String translation = new StringService(dto).getString();
+        if (StringUtil.isEmpty(translation)) {
+            return "";
+        }
+
+        if (args != null) {
```
jessiejuachon, line 199:
```suggestion
        if (!StringUtil.isEmpty(translation) && args != null) {
```

PR #715, opened 2020-08-13
use specific vip configuration for sample app

surite:


surite:
Found another root cause of sample app failure.

PR #735, opened 2020-08-21
G11n java client - Sample application with shared library

jessiejuachon:


linr211 (approved):

sample-client-app/sample-app-with-shared-lib/README.md
```diff
@@ -0,0 +1,36 @@
+Singleton Java Client Sample Application with Shared Library
+============
+
+This application can be used to demonstrate usage of a shared library that uses the Singleton Java client.
```
huihuiw01, line 4: demonstrate usage of a shared library 
may should be:
demonstrate usage of an application using a shared library

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/Main.java
```diff
@@ -0,0 +1,46 @@
...
+		} catch (VIPClientInitException e) {
+			System.out.println(e.getMessage());
+		}
+		cfg.initializeVIPService();
+		cfg.createTranslationCache(MessageCache.class);
+		cfg.createFormattingCache(FormattingCache.class);
+
+		TranslationMessage tm = (TranslationMessage) I18nFactory.getInstance(cfg).getMessageInstance(TranslationMessage.class);
+		String rootMsg = tm.getMessage(Locale.forLanguageTag("fr"), "default", "global_text_username");
+		System.out.println("Root application's message; fetch is triggered from the root application:  " + rootMsg);
```
huihuiw01, line 41: In actual application, initialize code(line 28-37) usually is not put together(put in one method) with getMessage code(line 39-41). Usually the former is called only once when application start, and the latter is called many times at where it's needed. 
So I think put initialize code in a method, and get message code in another method is better.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.vmware.vipclient.i18n.I18nFactory;
+import com.vmware.vipclient.i18n.VIPCfg;
+import com.vmware.vipclient.i18n.base.cache.FormattingCache;
+import com.vmware.vipclient.i18n.base.cache.MessageCache;
+import com.vmware.vipclient.i18n.base.instances.TranslationMessage;
+import com.vmware.vipclient.i18n.exceptions.VIPClientInitException;
+import com.vmware.vipclient.sample.shared.Shared;
+
+public class SharedLibDemo {
```
huihuiw01, line 17: This class is for showing how write code in shared library, so shouldn't this class put into sample-shared-library? 

Besides if it is not put into shared library, each application using shared library need write these code snippet.

sample-client-app/sample-shared-library/src/main/resources/sharedlib.properties
```diff
@@ -0,0 +1,27 @@
+# It's used for categorizing translation by product on Singleton side to maintain translation.
```
huihuiw01, line 1: Looks like this file is not used.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
```diff
@@ -0,0 +1,40 @@
...
+        sharedLibCfg.createTranslationCache(MessageCache.class);
+        sharedLibCfg.createFormattingCache(FormattingCache.class);
+
+        TranslationMessage tmSharedLib = (TranslationMessage) I18nFactory.getInstance().getMessageInstance(TranslationMessage.class, sharedLibCfg);
+
+        Shared shared = new Shared(tmSharedLib, locale);
+        System.out.println("Shared library's message; fetch triggered from inside the shared library: " + shared.someMethod());
+
+        String sharedMsg = tmSharedLib.getMessage(locale, "shared", "shared.library.key1");
+        System.out.println("Shared library's message; fetch is triggered from root application:  " + sharedMsg);
```
huihuiw01, line 37: Same comment as that in Main.java

huihuiw01 (changes requested):

sample-client-app/sample-app-with-shared-lib/README.md
jessiejuachon: No, what I put is what I actually meant.. to. demonstrate usage of (how to use) a shared library that uses the Singleton Java client

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/Main.java
jessiejuachon: We can do this as future improvement.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
jessiejuachon: This is for showing how to use a shared library's class and method (Shared.someMethod in lines 33-34) in a root application code such as SharedLibDemo.

Each root application that uses the shared library will call shared.someMethod() as usual (line 34). Only those root applications that also want to use VIP will have to write lines 19-33.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
jessiejuachon: We can do this as future improvement.

sample-client-app/sample-shared-library/src/main/resources/sharedlib.properties
jessiejuachon: This file will be used if sample-shared-library is used as a standalone app so let's keep it there.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
huihuiw01: My description maybe not precise, but the precondition I based on is both the root application and shared library use VIP, or there is no need for us to write these demo code.

The goal I want to reach is also let the root application call the shared library code as usual(before use VIP), and no big change for shared library.

But if put below code snippet in the root application, and root application call many methods of shared library, you need pass 'tmSharedLib' to each constructor of shared lib class called by the root application as line 31~33,
       TranslationMessage tmSharedLib = (TranslationMessage)  I18nFactory.getInstance().getMessageInstance(TranslationMessage.class, sharedLibCfg);  
        Shared shared = new Shared(tmSharedLib, locale);
        System.out.println("Shared library's message; fetch triggered from inside the shared library: " + shared.someMethod());

 this not only need add constructor in shared library, but also not call shared lib methods as usual, because before use VIP it call shared lib may like this:
        Shared shared = new Shared();
        System.out.println("Shared library's message; fetch triggered from inside the shared library: " + shared.someMethod());

But if you move these codes into sample-shared-library,  no need add constructor in shared lib and can call share lib methods as usual.

jessiejuachon: Yes, this sample is for the current implementation, which I did not design or implement. I am merely adding a sample usage through this PR. Improving the implementation can be in another PR.

jessiejuachon: Actually, Shared.java may act as a singleton instance that holds tmSharedLib which is to be used for the rest of the shared library classes. The point is to demonstrate how to pass the tmSharedLib, not how to design the shared lib.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/Main.java
jessiejuachon: The point is to demonstrate how to pass the tmSharedLib in the simplest way possible, not how to design the root app or the shared lib.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
jessiejuachon: The point is to demonstrate how to pass the tmSharedLib in the simplest way possible, not how to design the root app or the shared lib.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
huihuiw01: Yeah, Shared.java may act as a singleton instance that holds tmSharedLib which is to be used for the rest of the shared library classes. good idea!

But in this case Shared class should be called by shared lib's methods which need call getMessage to get translation, not by the root application, because the methods root application need call must not all in the Shared.java.

I think we should design sample app to make it easiest to use and as close to the real application as possible. So suggest future improvement.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
huihuiw01: I think we should design sample app to make it easiest to use and as close to the real application as possible. So suggest future improvement.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/Main.java
huihuiw01: I think we should design sample app to make it easiest to use and as close to the real application as possible. So suggest future improvement.

huihuiw01 (approved):

huihuiw01:
Since Jessiely will be on PTO starting tomorrow until the rest of the week. It would be good to have these PRs merged before her PTO, so that they can be tested while she is gone. So approve this PR.

But it's better do some improvement

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/Main.java
jessiejuachon: Yes, I agree. Actually, I think we should review the current design of shared library feature first.

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
jessiejuachon: @huihuiw01 , https://github.com/vmware/singleton/issues/744

sample-client-app/sample-app-with-shared-lib/src/main/java/com/vmware/vipclient/sample/SharedLibDemo.java
jessiejuachon: https://github.com/vmware/singleton/issues/744

Issue #717, opened 2020-08-14
[BUG] g11n-java-client TranslationMessage.getMessage for non-supported locales always triggers a request to VIP service  in a separate thread 


jessiejuachon:
**Describe the bug**
For non-supported locales, the cached fallback locale messages are returned immediately. This is the correct behavior. However, a separate thread is simultaneously launched to try to fetch the messages for the requested locale from the  VIP service. This done so that at runtime, in case support for the requested locale is added, the client application will be able to get the messages from VIP service. However, the thread is launched for every request, which does not change the result, but may be expensive in terms of application resource.

**To Reproduce**
1. Send request for a non-supported locale. Notice that a request to VIP is triggered every time.

**Expected behavior**
Requested locale must use the same maxAgeMillis as the fallback locale, but must have its own timestamp (and therefore, its own expiration time).
Send request to VIP service only after cache for the requested locale has expired.

lyiyu66:
The issue has been fixed with commit e8c3dde63ae25f36f967cd89435cbf26ef7457e5. Hence closing.

Issue #664, opened 2020-07-13
[BUG] [Java Client]When getMessages() returns to default locale's translation, cache can't get updated translation(default locale) after it is expired.


lyiyu66:
**Describe the bug**
Online or Mixed mode: When getMessages() returns to service default locale's translation, cache can't get updated translation(default locale) after it is expired.

**To Reproduce**
1. Set config with pure online or mixed mode.
2. Request a locale doesn't have its translation in service with getMessages(locale, component), then it will return default locale's translation from service.
3. Get one key's value from step2 returned result.
4. Update the default locale's value of that key and set the cache expired.
5. Repeat the request in step 2.
6. Repeat the request in step 2.
7. Repeat step 3.
Still get the original translation of default locale rather than the updated translation in step 4.

**Expected behavior**
Should get the updated translation of default locale once cache is expired and request is sent.

lyiyu66:
The issue also is reproducible when getMessages() return to default locale's translation from offlineBundles.

jessiejuachon:
(Pasting the same comment I put in issue #662 )

I just tested this and it works for me. I am guessing there is something wrong with the steps you wrote above. Please check the following:

In step 2, how is the client cache expire time set? If you are testing the new client workflow, make sure to NOT set the cacheExpiredTime property in the properties config file. Otherwise, cacheExpiredTime will be used instead of the value set by the Singleton service in the cache-control max age header in the HTTP response.
In step 3, which default locale message do you update? Since you are testing online or mixed mode, you should update the message in the service because that takes priority over the offline message.
In step 3, how do you expire the client-side cache?
If the client had previously cached the response and its expire time (either using the cacheExpiredTime config or the cache-control max age header), then the cached messages will only expire after the configured expire time. Do you wait for the time to pass, or do you have some other way to expire the time?

lyiyu66:
- I have set the cache expired time as 0l in config file to make cacheExpiredTime use new workflow(from http response header).
vipCfg.setCacheExpiredTime(0l);

- Depends on which default locale message is used. If it uses service default locale message(when service has requested component with default locale), then update the service side's, else(when service hasn't the requested component with default locale, but offlineBundles has), update offline message.

- Yes, I set cacahItem.setMaxAgeMillis(0l) or 1000l, then thread.sleep(2000), and also use cacheItem.isExpired() to check the expired status.

jessiejuachon:
[...]

Issue #662, opened 2020-07-13
[BUG] [Java Client]When getMessage() returns to default locale's translation, cache can't get updated translation(default locale) after it is expired.


lyiyu66:
**Describe the bug**
Online or Mixed mode: When getMessage() returns to service default locale's translation, cache can't get updated translation(default locale) after it is expired.

**To Reproduce**
1. Set config with pure online or mixed mode.
2. Request a locale doesn't have its translation in service with getMessage(locale, component, key), then it will return default locale's translation from service.
3. Update the default locale's value of this key and set the cache expired.
4. Repeat the request in step 2.
5. Repeat the request in step 2.
Still get the original translation of default locale rather than the updated translation in step 3. 

**Expected behavior**
Should get the updated translation of default locale once cache is expired and request is sent.

lyiyu66:
In mixed mode, the issue also is reproducible when getMessage() returns to default locale's translation from offlinebundles.

jessiejuachon:
Thanks, @lyiyu66 

How do you expire the cache in step 3?

lyiyu66:
> Thanks, @lyiyu66
> 
> How do you expire the cache in step 3? Is this the client-side cache?

Yes, the cache is client-side cache. Example as below:
Map<String, MessageCacheItem> m1 = ((MessageCache) c).getCachedTranslationMap();
m1.get("JavaClient_1.0.0_demo_false_#ru-CA").setMaxAgeMillis(0l);

jessiejuachon:
@lyiyu66 , this is ready for testing. One detail to change un your script: Call tm.getMessage(<default locale>,..) before tm.getMessage(<requested locale>,..). Only the call to the expired locale is able to update the expired locale's cache. 
The requested locale can just read the default locale's cache in case of failure, but not trigger any update for the default locale.

jessiejuachon:
@lyiyu66 , this is now ready to test

surite:
Test pass on commit 3b6f338c9c06dd6457baf6271eb2c21f7fed8aeb

PR #738, opened 2020-08-25
[g11n-java-client] Fix for issue #717 - non-supported locales trigger a request to VIP service in a separate thread for every getMessage/getMessages request

jessiejuachon:


Xiaochao8 (changes requested):
CI fails, could you check/

jessiejuachon:
> CI fails, could you check/

Seems like the new license header check is in place even if your PR that updates the year hasn't been merged? Anyway, I have updated for those 2 files.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -105,14 +105,19 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+    		if (cacheItem.getCachedData().isEmpty()) { // This means that the data to be used is from a fallback locale.
+				// If expired, try to first create and store cacheItem for the requested locale in a separate thread.
+				if (cacheItem.isExpired())
+					this.createCacheItemTask(fallbackLocalesIter);
+
+    			// Use cached data of cacheItem.getLocale() --> fallback locale
+				MessagesDTO fallbackLocaleDTO = new MessagesDTO(dto.getComponent(), cacheItem.getLocale(), dto.getProductID(), dto.getVersion());
+				cacheItem = new ComponentService(fallbackLocaleDTO).getMessages(null);
+			}
+			if (cacheItem.isExpired()) {
```
Xiaochao8, line 117: Should add an 'else' before if?

huihuiw01 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: No, we still want to refresh expired cacheItem even if it is for the fallback locale.

jessiejuachon: This is actually not new. It is the fix for https://github.com/vmware/singleton/issues/664 and https://github.com/vmware/singleton/issues/662

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -105,14 +105,19 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
     	MessageCacheItem cacheItem = null;
     	if (cacheService.isContainComponent()) { // Item is in cache
     		cacheItem = cacheService.getCacheOfComponent();
-    		if (cacheItem.isExpired()) { // cacheItem has expired
+    		if (cacheItem.getCachedData().isEmpty()) { // This means that the data to be used is from a fallback locale.
+				// If expired, try to first create and store cacheItem for the requested locale in a separate thread.
+				if (cacheItem.isExpired())
+					this.createCacheItemTask(fallbackLocalesIter);
```
Xiaochao8, line 111: For creating second time, it doesn't need to do fallback. How about using 'resreshCache' without locale fallback? This will improve performance.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
Xiaochao8: I think it's necessary to add it because the cache of fallback locale will be refreshed in line 115 because of the call getMessages with default locale. 
If no 'else', it will be refreshed twice, right?

Xiaochao8 (approved):

huihuiw01 (approved):

PR #753, opened 2020-08-31
G11n java client - Fix for issue#746 - MsgOriginsQueue cannot be cleared once initializing VIPCfg

jessiejuachon:


linr211 (approved):

Xiaochao8 (approved):

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -149,8 +149,9 @@ public void initialize(String cfg) throws VIPClientInitException {
         if (prop.containsKey("version"))
             this.version = prop.getString("version");
         
-        // Remote VIP resources takes priority over offline resources
+        // Remote VIP resources take priority over offline resources
         // so add DataSourceEnum.VIP first to msgOriginsQueue
+        this.setMsgOriginsQueue(new LinkedList<DataSourceEnum>());
```
huihuiw01, line 154: Is it better use 'this.msgOriginsQueue.clear();'?

jessiejuachon: I will use .clear inside the setter method because: 1. Good programming practice is to always use the setter method to change the private variable, even inside the same class. 2. The setter method can be set to a non-empty list later on, so we should just pass an empty list if we want to clear.

PR #774, opened 2020-09-11
use travis build matrix to test in parallel

surite:
execute following 4 tests parallel in separated environment:
1.check header
2.unit test
3.code scan

Benefits:
1.less time cost;
2.contributor can get all failures of all test in one time.(before, we use stages, the following stages will not be ran if current stage is failed)


WisdomLuo (dismissed):

surite:
travis hang, will re-create PR.

PR #782, opened 2020-09-14
Fix for bug 781 - ComponentService.getMessages is not retrieving "matchedLocale" from cache properly

jessiejuachon:
…ncorrectly using L2's data.

### DIFF
--- a/src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
+++ b/src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
@@ -21,6 +21,7 @@ public class MessageCacheItem implements CacheItem {
 	}
 	
 	public MessageCacheItem (String locale, Map<String, String> dataMap, String etag, long timestamp, Long maxAgeMillis) {
+		this.maxAgeMillis = maxAgeMillis == null ? this.maxAgeMillis : maxAgeMillis;
 		this.setCacheItem(locale, dataMap, etag, timestamp, maxAgeMillis);
 	}
 
--- a/src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
+++ b/src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
@@ -81,7 +81,6 @@ public class CacheService {
         Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
         if (c == null)
             return null;
-        CacheItem cacheItem = c.get(key);
-        return cacheItem;
+        return c.get(key);
     }
 }
--- a/src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
+++ b/src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
@@ -7,10 +7,11 @@ package com.vmware.vipclient.i18n.messages.service;
 import com.vmware.vipclient.i18n.VIPCfg;
 import com.vmware.vipclient.i18n.base.cache.Cache;
 import com.vmware.vipclient.i18n.base.cache.CacheItem;
-import com.vmware.vipclient.i18n.base.cache.FormatCacheItem;
 import com.vmware.vipclient.i18n.base.cache.MessageCacheItem;
+import com.vmware.vipclient.i18n.messages.dto.BaseDTO;
 import com.vmware.vipclient.i18n.messages.dto.MessagesDTO;
 import com.vmware.vipclient.i18n.util.ConstantsKeys;
+import com.vmware.vipclient.i18n.util.JSONUtils;
 import com.vmware.vipclient.i18n.util.LocaleUtility;
 
 import java.util.*;
@@ -23,11 +24,8 @@ public class CacheService {
     }
     
     public MessageCacheItem getCacheOfComponent(List<Locale> supportedLocales) {
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c == null)
-            return null;
         String cacheKey = dto.getCompositStrAsCacheKey();
-        MessageCacheItem cacheItem = (MessageCacheItem) c.get(cacheKey);
+        MessageCacheItem cacheItem = (MessageCacheItem) this.getCacheItem(cacheKey);
         if (cacheItem != null)
             return cacheItem;
         Locale matchedLocale = LocaleUtility.pickupLocaleFromList(supportedLocales,
@@ -35,74 +33,53 @@ public class CacheService {
         cacheKey = cacheKey.substring(0,
                 cacheKey.indexOf(ConstantsKeys.UNDERLINE_POUND) + 2)
                 + matchedLocale.toLanguageTag();
-        return (MessageCacheItem) c.get(cacheKey);
+        return (MessageCacheItem) this.getCacheItem(cacheKey);
     }
 
     public void addCacheOfComponent(MessageCacheItem itemToCache) {
         String cacheKey = dto.getCompositStrAsCacheKey();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c != null) {
-            c.put(cacheKey, itemToCache);
-        }
+        this.addCacheItem(cacheKey, itemToCache);
     }
 
     public void updateCacheOfComponent(MessageCacheItem itemToCache) {
         String cacheKey = dto.getCompositStrAsCacheKey();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c != null) {
-            MessageCacheItem cacheItem = (MessageCacheItem) c.get(cacheKey);
-            if (cacheItem == null) {
-            	cacheItem = new MessageCacheItem();
-            	c.put(cacheKey, cacheItem);
-            }
-            cacheItem.setCacheItem(itemToCache);
+        MessageCacheItem cacheItem = (MessageCacheItem) this.getCacheItem(cacheKey);
+        if (cacheItem == null) {
+            cacheItem = new MessageCacheItem();
+            this.addCacheItem(cacheKey, cacheItem);
         }
+        cacheItem.setCacheItem(itemToCache);
     }
 
     public boolean isContainComponent(List<Locale> supportedLocales) {
         String cacheKey = dto.getCompositStrAsCacheKey();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c == null)
-            return false;
-
-        if (c.get(cacheKey) != null)
+        if (this.getCacheItem(cacheKey) != null)
             return true;
         Locale matchedLocale = LocaleUtility.pickupLocaleFromList(supportedLocales,
                 this.getLocaleByCachedKey(cacheKey));
         cacheKey = cacheKey.substring(0,
                 cacheKey.indexOf(ConstantsKeys.UNDERLINE_POUND) + 2)
                 + matchedLocale.toLanguageTag();
-        return c.get(cacheKey) != null;
+        return this.getCacheItem(cacheKey) != null;
     }
 
     public boolean isContainStatus() {
-        boolean f = false;
         String cacheKey = dto.getTransStatusAsCacheKey();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c != null) {
-            f = c.keySet().contains(cacheKey);
-        }
-        return f;
+        return this.getCacheItem(cacheKey) != null;
     }
 
     public Map<String, String> getCacheOfStatus() {
         String cacheKey = dto.getTransStatusAsCacheKey();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c != null) {
-        	MessageCacheItem cacheItem = (MessageCacheItem) c.get(cacheKey);
-        	if (cacheItem != null) {
-        		return (Map<String, String>) cacheItem.getCachedData();
-        	}
+        MessageCacheItem cacheItem = (MessageCacheItem) this.getCacheItem(cacheKey);
+        if (cacheItem != null) {
+            return cacheItem.getCachedData();
         }
         return null;
     }
 
     public void addCacheOfStatus(Map<String, String> dataMap) {
         String cacheKey = dto.getTransStatusAsCacheKey();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c != null) {
-            c.put(cacheKey, new MessageCacheItem(dataMap));
-        }
+        addCacheItem(cacheKey, new MessageCacheItem(dataMap));
     }
 
     private Locale getLocaleByCachedKey(String key) {
@@ -110,5 +87,35 @@ public class CacheService {
                 key.indexOf(ConstantsKeys.UNDERLINE_POUND) + 2, key.length());
         return Locale.forLanguageTag(locale.replace("_", "-"));
     }
-    
+
+    public void addSupportedLanguages(BaseDTO dto, MessageCacheItem cacheItem) {
+        String cacheKey = getSupportedLanguagesCacheKey(dto);
+        addCacheItem(cacheKey, cacheItem);
+    }
+
+    public MessageCacheItem getSupportedLanguages(BaseDTO dto) {
+        String cacheKey = getSupportedLanguagesCacheKey(dto);
+        return (MessageCacheItem) getCacheItem(cacheKey);
+    }
+
+    private String getSupportedLanguagesCacheKey(BaseDTO dto){
+        return ConstantsKeys.DISPNS_PREFIX + dto.getProductID() + ConstantsKeys.UNDERLINE + dto.getVersion();
+    }
+
+    private void addCacheItem(String key, CacheItem cacheItem) {
+        if (key != null && cacheItem != null) {
+            Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
+            if (c != null) {
+                c.put(key, cacheItem);
+            }
+        }
+    }
+
+    private CacheItem getCacheItem(String key) {
+        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
+        if (c == null)
+            return null;
+        CacheItem cacheItem = c.get(key);
+        return cacheItem;
+    }
 }
--- a/src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
+++ b/src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
@@ -8,7 +8,6 @@ import com.vmware.vipclient.i18n.VIPCfg;
 import com.vmware.vipclient.i18n.base.cache.Cache;
 import com.vmware.vipclient.i18n.base.cache.CacheItem;
 import com.vmware.vipclient.i18n.base.cache.MessageCacheItem;
-import com.vmware.vipclient.i18n.messages.dto.BaseDTO;
 import com.vmware.vipclient.i18n.messages.dto.MessagesDTO;
 import com.vmware.vipclient.i18n.util.ConstantsKeys;
 import com.vmware.vipclient.i18n.util.LocaleUtility;
@@ -22,25 +21,8 @@ public class CacheService {
         this.dto = dto;
     }
 
-    private List<Locale> getLocalesOfCachedMsgs() {
-        List<Locale> locales = new LinkedList<>();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c != null) {
-            Set<String> cacheKeys = c.keySet();
-            for (String key : cacheKeys) {
-                Locale locale = getLocaleByCachedKey(key);
-                if (locale != null)
-                    locales.add(locale);
-            }
-        }
-        return locales;
-    }
-
     public MessageCacheItem getCacheOfComponent() {
         String cacheKey = dto.getCompositStrAsCacheKey();
-        MessageCacheItem cacheItem = (MessageCacheItem) this.getCacheItem(cacheKey);
-        if (cacheItem != null)
-            return cacheItem;
         Locale matchedLocale = LocaleUtility.pickupLocaleFromList(this.getLocalesOfCachedMsgs(),
                 this.getLocaleByCachedKey(cacheKey));
         cacheKey = cacheKey.substring(0,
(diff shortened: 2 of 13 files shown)


src/test/java/com/vmware/vipclient/i18n/messages/service/CacheServiceTest.java
```diff
@@ -225,4 +229,26 @@ public void testExpireUsingCacheExpiredTimeConfig() throws InterruptedException
...
+        cfg.initializeMessageCache();
+
+        CacheService cs = new CacheService(new MessagesDTO());
+        List<Locale> supportedLocales = cs.getSupportedLocalesFromCache();
+        assertTrue(supportedLocales.contains(Locale.forLanguageTag("fil")));
+        assertEquals("Filipino", supportedLocales.get(
+                supportedLocales.indexOf(Locale.forLanguageTag("fil"))).getDisplayName());
+
+        // Disable offline mode off for next tests.
+        cfg.setOfflineResourcesBaseUrl(offlineResourcesBaseUrlOrig);
```
Xiaochao8, line 251: To make sure these 2 lines can be performed, above code should be in catch block.

huihuiw01:
Hi Jessiely,
     I'm curious about the impact without the change in this PR, could you explain it in the issue description?

     In my opinion, it's correct to get supported locales by calling L2 API; it's incorrect to get supported locales from L3's cache, because you also add not supported locales to L3 cache, though the content is that of fallback locale.

src/test/java/com/vmware/vipclient/i18n/messages/service/CacheServiceTest.java
```diff
@@ -238,17 +238,19 @@ public void testGetSupportedLocalesOfflineBundles() {
...
 
-        // Disable offline mode off for next tests.
-        cfg.setOfflineResourcesBaseUrl(offlineResourcesBaseUrlOrig);
-        cfg.setMsgOriginsQueue(msgOriginsQueueOrig);
+            CacheService cs = new CacheService(new MessagesDTO());
+            List<Locale> supportedLocales = cs.getCachedLocales();
+            assertTrue(supportedLocales.contains(Locale.forLanguageTag("fil")));
+            assertEquals("Filipino", supportedLocales.get(
+                    supportedLocales.indexOf(Locale.forLanguageTag("fil"))).getDisplayName());
+        } catch (Exception e) {
```
Xiaochao8, line 250:
```suggestion
        } finally {
```

jessiejuachon:
> Hi Jessiely,
> I'm curious about the impact without the change in this PR, could you explain it in the issue description?
> 
> ```
>      In my opinion, it's correct to get supported locales by calling L2 API; it's incorrect to get supported locales from L3's cache, because you also add not supported locales to L3 cache, though the content is that of fallback locale.
> ```

Hi Huihui. If you look at how the method is used, it is used not to get the list of supported locales. It is used to get the list of already cached locales. That is also why I changed the name of it.

jessiejuachon:
> Hi Jessiely,
> I'm curious about the impact without the change in this PR, could you explain it in the issue description?
> 
> ```
>      In my opinion, it's correct to get supported locales by calling L2 API; it's incorrect to get supported locales from L3's cache, because you also add not supported locales to L3 cache, though the content is that of fallback locale.
> ```

Please see new changes. We need a separate cacheItem entry for this one because later on, it needs to have its own expire time. When expired, need to fetch the list from data store again in case a new locale has been recently added to the VIP service. Hence, we should not use just any 1 of the displaylocale-specific cache items.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -25,19 +22,20 @@ public CacheService(MessagesDTO dto) {
...
         cacheKey = cacheKey.substring(0,
                 cacheKey.indexOf(ConstantsKeys.UNDERLINE_POUND) + 2)
                 + matchedLocale.toLanguageTag();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c == null) {
-            return null;
-        } else {
-            return (MessageCacheItem) c.get(cacheKey);
-        }
+        return (MessageCacheItem) c.get(cacheKey);
```
Xiaochao8: What's this change for?

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -101,17 +99,14 @@ public void addCacheOfStatus(Map<String, String> dataMap) {
...
-        List<Locale> locales = new ArrayList<>();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c == null) {
-            return locales;
-        }
-        Set<String> cacheKeys = c.keySet();
-        for (String key: cacheKeys) {
-            locales.add(getLocaleByCachedKey(key));
+    public List<Locale> getSupportedLocalesFromCache() {
+        List<Locale> result = new LinkedList<>();
```
Xiaochao8, line 103: Product service should consume cache service. But here cache service makes use of product service. It seems a design problem.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -101,17 +99,14 @@ public void addCacheOfStatus(Map<String, String> dataMap) {
...
-    public List<Locale> getCachedLocales() {
-        List<Locale> locales = new ArrayList<>();
-        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
-        if (c == null) {
-            return locales;
-        }
-        Set<String> cacheKeys = c.keySet();
-        for (String key: cacheKeys) {
-            locales.add(getLocaleByCachedKey(key));
+    public List<Locale> getSupportedLocalesFromCache() {
```
Xiaochao8, line 102: Don't know how the locale are from cache. Does product service get data from cache?

jessiejuachon: If you see in ProductService, it uses the cache now.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
jessiejuachon: You have a point. Let me move things around.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -77,18 +73,31 @@ public ProductService(BaseDTO dto) {
...
-            if (locales == null || locales.isEmpty()) {
+            if (locales.isEmpty()) {
                 logger.error(ConstantsMsg.GET_LOCALES_FAILED, dataSource.toString());
             }
         }
+
+        // Add list of supported locales to cache
+        Map<String, String> cacheMap = new HashMap<>();
+        for (String locale : locales) {
+            cacheMap.put(locale, "");
```
huihuiw01, line 97: I think you needn't convert locale list to locale map, just add locale list to cache is enough.

jessiejuachon: FormattingCacheItem has a Map<String,String> property. How do you add a list instead?

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
jessiejuachon: See issue 781 for description.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
huihuiw01: @xiaochao8 I think it's ok to call product service here, because it's reasonable for service classes to call each other.
To avoid this, I see Jessiely need make changes to multiple files,  I think it's unnecessary. How do you think?

@jessiejuachon Not sure if it's convenient to revert the change for this?

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
huihuiw01: Get it, it make sense to convert it to a map for reusing cacheItem Object.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
jessiejuachon: @Xiaochao8 , @huihuiw01 . I think you both have a point. 

Huihui, yes it should be reasonable for service classes to call each other. If you see the most recent changes. I call ProductService inside ComponentService. However, I think Xiaochao's point is more about using cache services (CacheService or FormattingCacheService) only for cache-related operations. 

I was calling ProductService.getSupportedLocales inside CacheService.  ProductService.getSupportedLocales is not just doing cache-related operation. If nothing is found in cache, it tries to retrieve the data from data store. It would be incorrect to call this method inside a CacheService method--it would be misleading as CacheService should be for cache-related operations only.

jessiejuachon: Actually, I realized that what's needed inside CacheService.isContainComponent and CacheService.getCacheOfComponent is not the actual list of supported locales. Instead, it only needs the list of locales of currently cached messages. On other words, if a locale may supported in VIP service, but its messages are not yet in cache. Which means it should not be included in the list to pick from.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
Xiaochao8: If it is the cases. This method should belong to product service. This is the normal way to get locales. get locale request -> service -> cache-> data source.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -6,127 +6,101 @@
...
+            }
+        }
+    }
+
+    private CacheItem getCacheItem(String key) {
+        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
+        if (c == null)
+            return null;
+        CacheItem cacheItem = c.get(key);
+        return cacheItem;
```
Xiaochao8, line 104:
```suggestion
        return c.get(key);
```

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -6,127 +6,101 @@
...
-            }
-            cacheItem.setCacheItem(itemToCache);
+        MessageCacheItem cacheItem = (MessageCacheItem) this.getCacheItem(cacheKey);
+        if (cacheItem == null) {
+            cacheItem = new MessageCacheItem();
+            this.addCacheItem(cacheKey, cacheItem);
         }
-    }
-
-    public boolean isContainComponent() {
```
Xiaochao8, line 65: Please don't remove this method, this is to fix a bug about source collection. Need to store null objects to cache to avoid bugs. 
You can confirm this with Colin.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -6,127 +6,101 @@
...
+                Locale locale = getLocaleByCachedKey(key);
+                if (locale != null)
+                    locales.add(locale);
+            }
         }
         return locales;
     }
 
     private Locale getLocaleByCachedKey(String key) {
+        if (key.startsWith(ConstantsKeys.DISPNS_PREFIX))
```
Xiaochao8, line 83: We should have a prefix for translation bundles. If not and  there is a new prefix added, this line needs to be changed.

Xiaochao8 (commented):
Because of so many changes, are all the changes necessary to fix the bug?

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
jessiejuachon: This change is not related to this PR so I have removed it.

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
jessiejuachon: I put it back, but if you check the code, it is exactly the same as `CacheService.getCacheOfComponent()!=null`

On that note, we are not going to use the java client library for source collection anymore, even for existing products.
Also, if you check where CacheService.addCacheOfComponent is invoked, you will see that we do not store null in the cache. We do not even store a MessageCacheItem with an empty map.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -91,4 +87,18 @@ public ProductService(BaseDTO dto) {
         }
         return locales;
     }
+
+    public Set<Locale> getSupportedLocales() {
```
Xiaochao8, line 91:
```suggestion
    public Set<Locale> getSupportedLocaleSet() {
```

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -77,7 +73,7 @@ public ProductService(BaseDTO dto) {
      *
      * @return list of locales of the product specified in the dto object
      */
-    public List<String> getSupportedLocales(){
+    public List<String> getSupportedLanguageTags(){
```
Xiaochao8, line 76: Suggest not to change this method but to name new method to another name.
```suggestion
    public List<String> getSupportedlocales(){
```

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
jessiejuachon: @Xiaochao8 , did you check where this method is used? Do you agree that we can remove it to simplify code?

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: We actually only need 1 method. I will use: `public Set<Locale> getSupportedLocales() `to replace the following 2 signatures:

- public Set<Locale> getSupportedLocales()
- public List<String> getSupportedLocales()

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: See my comment above.

jessiejuachon:
> Because of so many changes, are all the changes necessary to fix the bug?

@Xiaochao8 , I am not sure if this comment was before or after I rolled back most changes. Do you agree that the changes right now are minimal? This PR now has only 7 files changed, 3 of those are test files. Among the 4 other files, the main code logic is in ComponentService.createCacheItem (lines 135-145). The other changes are mostly code clean up.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -130,11 +131,23 @@ private MessageCacheItem createCacheItem(Iterator<Locale> fallbackLocalesIter) {
...
+		if (!ps.isSupportedLocale(Locale.forLanguageTag(dto.getLocale()))) {
+			Locale matchedLocale = LocaleUtility.pickupLocaleFromList(new LinkedList<>(ps.getSupportedLocales()), Locale.forLanguageTag(dto.getLocale()));
+			if (ps.isSupportedLocale(matchedLocale)) {
+				MessagesDTO matchedLocaleDTO = new MessagesDTO(dto.getComponent(), matchedLocale.toLanguageTag(), dto.getProductID(), dto.getVersion());
+				cacheItem = new ComponentService(matchedLocaleDTO).getMessages(null);
+				MessageCacheItem cacheItemCopy = new MessageCacheItem(matchedLocale.toLanguageTag(), null, null, System.currentTimeMillis(), cacheItem.getMaxAgeMillis());
+				cacheService.addCacheOfComponent(cacheItemCopy);
+				return cacheItem;
+			}
+		}
```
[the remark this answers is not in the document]
jessiejuachon: @huihuiw01 , you are missing some use cases in the above code.
1. We want fallbackLocalesIter to be a parameter of the method.  If you trace, getMessages eventually calls createCacheItem if cahceItem is not yet in cache. In there, if fallbackLocalesIter is null, no fallback is applied. We need this for use cases where fallback is not necessary.
2. A "matched locale" does not necessarily mean that it is supported. For example: supported locales may include "fr" and "en_US". If the requested locale is "fr_CA", it matches "fr", but fr_CA" is technically NOT supported. For this, the client will use "fr" bundle only because fr_CA bundle does not exist. If the service is updated to support fr_CA (i.e. supported locales are now "fr", "en_US" and "fr_CA"), then the client must not use the locale matching logic so that it will pick up "fr_CA" from the service.

That being said, you are right, we do not need to cache non-supported locales with an empty map anymore because we are now using getSupportedLocales in here. However, if you look at the implementation of ProductService.getSupportedLocales, it is not yet using the cache store this list. Hence, every call to it is expensive. So for now, the call to it is in createCacheItem only. I plan to update this in PR754 which is planned for next Sprint.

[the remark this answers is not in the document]
jessiejuachon: @huihuiw01 , you are missing a lot of use cases in the above code.

1. We need to create a new CacheItem if it is not found in cache (line 119).
2. We need to check for expiry of MessageCacheItem objects and refresh the content accordingly. These are in lines 108-109, 116
3. We want fallbackLocalesIter to be a parameter of the method. If you trace, the getMessages method eventually calls createCacheItem (if cacheItem is not yet in cache). In there, if fallbackLocalesIter is null, no fallback is applied. We need this for use cases where fallback is not necessary.
4. A "matched locale" does not necessarily mean that it is "supported". For example: supported locales may include "fr" and "en_US". If the requested locale is "fr_CA", it matches "fr", but fr_CA" is technically NOT supported. For this, the client will use "fr" bundle only because "fr_CA" bundle does not exist. Later on, if the service is updated to support "fr_CA" (i.e. supported locales are now "fr", "en_US" and "fr_CA"), then the client must not use the locale matching logic so that it will pick up "fr_CA" from the service. That is why there is another "if" in line 138.
5. There is a reason why I am caching non-supported locales with an empty cacheData map instead of directly serving the next available fallback. We want the client to eventually check the service if the non-supported locale has been made available in the service. In other words, if the locale is now supported. The only way to do this is to add an "expiry" that is specific to the cached non-supported locale so that when it expires, it means it is time to check again from the VIP service. Note: We can not check from VIP service for every request due to performance impact so we wait for expiry of the non-supported locale. This is how it is implemented:

For a non-supported locale, a MessageCacheItem with an empty cacheData map, the current timestamp, and the maxAgeMillis of the fallback locale is created. This MessageCacheItem is added to cache using the non-supported locale in the cache key. This logic is in line 157-158.
Now when this MessageCacheItem is accessed, the cacheData map is empty and so the corresponding fallback locale's data map has to be served (lines 112 -113). 

I believe #5 is very much related to my cache expiry logic comments in your PR. Do you need to connect over a phone call to further clarify?
https://github.com/vmware/singleton/pull/742#discussion_r482185125
https://github.com/vmware/singleton/pull/742#discussion_r482210946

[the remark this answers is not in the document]
jessiejuachon: @huihuiw01 , you are missing some use cases in your code suggestion above.

1. We need to fetch data from data store and create a new CacheItem if data is not found in cache (see line 119).
2. We need to check for expiry of MessageCacheItem objects and refresh the content accordingly. These are in lines 108-109, 116
3. We want fallbackLocalesIter to be a parameter of the method. If you trace getMessages, it eventually calls createCacheItem (if cacheItem is not yet in cache). In createCacheItem, if fallbackLocalesIter is null, no fallback is applied. We need this for use cases where fallback is not necessary.

That being said, yes we do not need to cache non-supported locales with an empty cacheData map anymore because we are now using ProductService.getSupportedLocales to identify which is supported and which is not. However, if you look at ProductService.getSupportedLocales, it is not yet using the cache, so each call to this method is expensive (as of now). Hence, I was planning to postpone code clean up to 757.  What do you think?

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
```diff
@@ -21,6 +21,7 @@ public MessageCacheItem(Map<String, String> dataMap) {
 	}
 	
 	public MessageCacheItem (String locale, Map<String, String> dataMap, String etag, long timestamp, Long maxAgeMillis) {
+		this.maxAgeMillis = maxAgeMillis == null ? this.maxAgeMillis : maxAgeMillis;
```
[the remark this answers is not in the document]
jessiejuachon: No, just code clean up. We do not want to override the default value with a null. Need separate PR?

src/main/java/com/vmware/vipclient/i18n/messages/service/CacheService.java
```diff
@@ -6,60 +6,37 @@
```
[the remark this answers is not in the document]
jessiejuachon: It is just code clean up. The following were repeated everywhere so I put in 2 separate methods:

` private void addCacheItem(String key, CacheItem cacheItem) {
        if (key != null && cacheItem != null) {
            Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
            if (c != null) {
                c.put(key, cacheItem);
            }
        }
    }

    private CacheItem getCacheItem(String key) {
        Cache c = VIPCfg.getInstance().getCacheManager().getCache(VIPCfg.CACHE_L3);
        if (c == null)
            return null;
        return c.get(key);
    }`

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -130,11 +131,28 @@ private MessageCacheItem createCacheItem(Iterator<Locale> fallbackLocalesIter) {
...
```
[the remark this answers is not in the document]
jessiejuachon: Zoom call please

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -130,11 +130,28 @@ private MessageCacheItem createCacheItem(Iterator<Locale> fallbackLocalesIter) {
...
 		// Create a new cacheItem object to be stored in cache
 		MessageCacheItem cacheItem = new MessageCacheItem();
-		refreshCacheItem(cacheItem, VIPCfg.getInstance().getMsgOriginsQueue().iterator());
 
+		// If the requested locale is not supported, but matches a supported locale (eg. requested locale "fr_CA matches supported locale "fr"),
+		// return the messages of the supported locale that best matches the requested locale.
+		ProductService ps = new ProductService(dto);
+		if (!ps.isSupportedLocale(Locale.forLanguageTag(dto.getLocale()))) {
+			Locale matchedLocale = LocaleUtility.pickupLocaleFromList(new LinkedList<>(ps.getSupportedLocales()), Locale.forLanguageTag(dto.getLocale()));
+			if (!matchedLocale.equals(Locale.forLanguageTag(dto.getLocale()))) {
```
huihuiw01, line 139: I think this judgement may not needed, because no matter there is matched locale or not, below code need be executed. 
for example, fr supported, no matter request fr-CA or ru, below code should be executed.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -130,11 +130,28 @@ private MessageCacheItem createCacheItem(Iterator<Locale> fallbackLocalesIter) {
...
+				MessagesDTO matchedLocaleDTO = new MessagesDTO(dto.getComponent(), matchedLocale.toLanguageTag(), dto.getProductID(), dto.getVersion());
+				cacheItem = new ComponentService(matchedLocaleDTO).getMessages(null);
+				MessageCacheItem cacheItemCopy = new MessageCacheItem(matchedLocale.toLanguageTag(), null, null, System.currentTimeMillis(), cacheItem.getMaxAgeMillis());
+				cacheService.addCacheOfComponent(cacheItemCopy);
+				return cacheItem;
+			}
+		}
```
huihuiw01, line 146: I think here still lack else logic, when there is no matched locale, you need get fallback locale's content for request locale.
for example, only fr supported, when request for ru 
ru -> default locale (not found)-> source locale.

jessiejuachon: That happens in line 151. It will continue there without entering the "if".

huihuiw01 (approved):

Xiaochao8 (approved):

PR #784, opened 2020-09-14
Fix for issue 783 - When initializing the cache, offline only locales are not included. 

jessiejuachon:
… are not included.

### DIFF
--- a/src/test/java/com/vmware/vip/i18n/TranslationMessageTest.java
+++ b/src/test/java/com/vmware/vip/i18n/TranslationMessageTest.java
@@ -258,9 +258,9 @@ public class TranslationMessageTest extends BaseTestClass {
     }
 
     @Test
-    public void testGetAllComponentTranslation() {
+    public void testGetAllComponentTranslationOnlineModeOnly() {
         List<Map> list = new ProductService(dto).getAllComponentTranslation();
-        Assert.assertTrue(list.size() > 0);
+        Assert.assertTrue(list.size() == 24); // 8 locales
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/com/vmware/vip/i18n/TranslationMessageTest.java
```diff
@@ -258,26 +258,23 @@ public void testGetComponentMessages_() {
...
     }
 
     @Test
     public void testGetAllComponentTranslationMixedMode() {
         String offlineResourcesBaseUrlOrig = vipCfg.getOfflineResourcesBaseUrl();
         vipCfg.setOfflineResourcesBaseUrl("offlineBundles/");
         List<DataSourceEnum> msgOriginsQueueOrig = vipCfg.getMsgOriginsQueue();
-        vipCfg.setMsgOriginsQueue(new LinkedList<DataSourceEnum>(Arrays.asList(DataSourceEnum.VIP, DataSourceEnum.Bundle)));
-        String vipServerOrig = vipCfg.getVipServer();
-        vipCfg.setVipServer("http://1.1.1.1:80");
```
Xiaochao8, line 273: Why remove this line?

huihuiw01:
I still wonder if such case exsits 'locales only supported in offline mode', could you give me an example?

src/test/java/com/vmware/vip/i18n/TranslationMessageTest.java
jessiejuachon: Because we want the server hostname to be correct to enable online mode (the one that has a mocked response).

jessiejuachon:
> I still wonder if such case exsits 'locales only supported in offline mode', could you give me an example?

1. If a locale bundle was not loaded into the service (service-side bug).
2. If at the time of caching the "supported locales" (which is usually at application start up), the service bundle is not loaded yet (e.g. in the SaaS version of VIP).

huihuiw01:
> > I still wonder if such case exsits 'locales only supported in offline mode', could you give me an example?
> 
>     1. If a supposedly supported locale bundle was not loaded into the service (service-side bug).
>        At application start up, if initializeCache=true, the cache should be populated.  If the service bundle is not loaded into the service yet, offline bundles should be loaded.
>        For instance, SaaS product has just been deployed with the offline bundles. The cache will be populated at app start up (initializeCache=true), but SaaS version of VIP hasn't been updated yet. It should pick up the offline bundle/s.
> 
>     2. message_source.json is from offline if offline mode is supported, so it should be loaded in cache accordingly.
> 
> 
> The point is that the we now support locales from both online and offline locations (mixed mode). So in principle, the proper response should be a combination of both online and offline, even if the use case happens rarely.

1. If in this case, should component list be changed like this?

2. message_source.json only exists in offline bundle? But I see the code still fetch it from service side(refer to ComponentService), so I think it also exists in service side. If it only exists in offline bundle, should we  skip fetch it from service side?

jessiejuachon:
> > > I still wonder if such case exsits 'locales only supported in offline mode', could you give me an example?
> > 
> > 
> > ```
> > 1. If a supposedly supported locale bundle was not loaded into the service (service-side bug).
> >    At application start up, if initializeCache=true, the cache should be populated.  If the service bundle is not loaded into the service yet, offline bundles should be loaded.
> >    For instance, SaaS product has just been deployed with the offline bundles. The cache will be populated at app start up (initializeCache=true), but SaaS version of VIP hasn't been updated yet. It should pick up the offline bundle/s.
> > 
> > 2. message_source.json is from offline if offline mode is supported, so it should be loaded in cache accordingly.
> > ```
> > 
> > 
> > The point is that the we now support locales from both online and offline locations (mixed mode). So in principle, the proper response should be a combination of both online and offline, even if the use case happens rarely.
> 
> 1. If in this case, should component list be changed like this?
> 2. message_source.json only exists in offline bundle? But I see the code still fetch it from service side(refer to ComponentService), so I think it also exists in service side. If it only exists in offline bundle, should we  skip fetch it from service side?

You are right regarding component list. I will update this soon.

Sorry for not being clear with messages_source. Flow for "source" locale should be the same as in any other locale. Source messages also exist in service (a.k.a latest). Just like for any locale, we want to pick up the bundle from remote if it is available. If not available, we want to use the local messages_source.json. No special case here, so same as #1.

PR #805, opened 2020-09-23
G11n java client - Caching the list of supported locales used by L3 to enhance performance

jessiejuachon:


Xiaochao8 (dismissed):

Xiaochao8 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -71,28 +75,81 @@ public ProductService(BaseDTO dto) {
...
+        FutureTask<MessageCacheItem> task = new FutureTask<>(callable);
+        Thread thread = new Thread(task);
+        thread.start();
+    }
+
+    private MessageCacheItem createCacheItem (DataSourceEnum dataSource) {
+        CacheService cs = new CacheService(new MessagesDTO(dto));
+        MessageCacheItem cacheItem = new MessageCacheItem();
+        refreshCacheItem(cacheItem, dataSource);
+        if (!cacheItem.getCachedData().isEmpty()) {
```
huihuiw01, line 139: I think this judgement is not needed. No matter the locale list is empty or not, it should be cached, or it will always try to fetch locale list from data source if last fetch fails, which may impact the performance.

huihuiw01:
Does component list need be changed like this?

Xiaochao8 (approved):

huihuiw01 (approved):

Issue #758, opened 2020-09-06
[BUG] [Java Client]Pattern(Date time/Number/Currency/Percent) fail to fallback to source locale if requested and default locale can't be fetched successfully.


lyiyu66:
**Describe the bug**
Pattern(Date time/Number/Currency/Percent) fail to fallback to source locale if requested and default locale can't be fetched successfully.

**To Reproduce**
1. Set defaultlocale = invaild and sourcelocale = de in java config properties file.
2. Request a invalid locale(e.g. abc) to get patterns((Date time/Number/Currency/Percent).

**Expected behavior**
Should get de locale's patterns.

**Screenshots**
Example of get number pattern
![image](https://user-images.githubusercontent.com/51813046/92328520-fe911080-f093-11ea-887b-246a74c90159.png)

huihuiw01:
The reason for this issue is Jessiely required L2 use the same fallback locales chain as L3, and in the chain source locale is hardcode as 'source', doesn't use the source locale configed in the config file. 
But there are resource files suffixed with 'source' for L3, no such files for L2, so this bug occurred.

Need discuss if this fallback chain is appropriate to L2.

huihuiw01:
@jessiejuachon After discussion with Demin and team members in last Friday's sprint review meeting, the conclusion of the fallback queue for L2 is:

requested locale -> default locale -> English locale

As L2 has nothing to do with source locale.

jessiejuachon:
Proposed solution: Use LocaleUtility.getSourceLocale in L2 workflow whenever locale = "source" in locale fallback queue
Also, can we change priority to low?

lyiyu66:
Lower the priority to low as the corresponding data about default locale is always available.

jessiejuachon:
Discussed with Huihui and Xiaochao, then with Demin. Here's the conclusion:
Use the same locale fallback queue for L2 and L3. 

Using the source locale must not happen in production because default locale should be always available. If source locale is used instead of default, then it is a pre-deployment bug that g11n QE must report. 

Having said that, the bug that Yiyu filed is valid, but of low priority. This is because it is expected to always find the default locale for L2 in production so it shouldn't even reach source locale. Nonetheless, to solve the bug, we should use getSourceLocale for L2. 

Here is my proposed code changes to Huihui's PR: https://github.com/huihuiw01/singleton/pull/1/files

huihuiw01:
After discussion on Jan 22 2021, the final conclusion for L2 fallback is: 
[...]

PR #806, opened 2020-09-24
fix bug 758: fail to fallback to source locale when default locale is not found

huihuiw01:


src/main/java/com/vmware/vipclient/i18n/util/LocaleUtility.java
```diff
@@ -19,7 +19,8 @@
     
     // A locale fallback priority queue. For now, it only contains the default locale and the source messages (from "messages_source.json").
     private static List<Locale> fallbackLocales = new LinkedList<Locale>(Arrays.asList(getDefaultLocale(), Locale.forLanguageTag(ConstantsKeys.SOURCE)));
-    
+    private static List<Locale> l2FallbackLocales = new LinkedList<Locale>(Arrays.asList(getDefaultLocale(), getSourceLocale()));
```
jessiejuachon, line 22: We do not need getSourceLocale() for L2. It is specifically for L3 purposes. "Source" means the locale of the original messages that were translated in multiple languages. It has nothing to do with L2.

Note: The method getSourceLocale must actually be updated to return the locale used in messages_source.json, instead of manually configuring it in vipconfig.properties.  -- This can be done in a future enhancement.

We only add "source" in the fallbackLocales queue because it means messages_source.json is bundled with the application code. This "messages_source,json"  is used only when it has new/updated L3 strings that are not yet in in the actual messages_xx.json of the source locale (a.k.a out of sync). This works for L2 just fine: L2 workflow must find a fallback locale in the queue before it reach "source". If it reaches "source", then it will fail also which is expected.

That being said, why do we need to maintain 2 separate queues? Aside form "source", will the content be  any different? It won't be good to have to maintain 2 queues.

Xiaochao8: I think source can be removed from the fallbacklocales because it's very special. Therer is no source in singleton service, even in L3 flow, many places need to handle it specially.
So remove it from fallbacklocales and when there is no any translation for the requested locale, get 'source' from local.

huihuiw01: I agree with xiaochao to remove source locale from locale fallback chain, especially for that of L2. 

If reuse the same locale fallback chain with L3, it may confuse/mislead other team members when they need maintain the code and look at the code at first insight that L2 need fallback to source locale.

Since L2 needn't fallback to source locale in theory, why keep the logic?

In addition, if keep fallback to source locale, in debug mode, the log will print message can't find source pattern which may mislead others the error is lack of source pattern file, actually source pattern is not needed for L2.

So I think we should either remove source locale from fallback queue if L2 and L3 share the same queue or keep standalone fallback queue for L2.

jessiejuachon: Regarding Xiaochao's comment:
First, there is no more discussion whether falling back to source messages is necessary for L3. It is an existing feature, even in the old client workflow, and we will continue to support it. The client library must return source message if source message has been recently updated (does not match the message in the messages_<locale>.json). We call this use case the "out-of-sync" stage where the source bundle does not match the translation bundles yet as the translation bundles are lagging behind.

Second, it is not true that source messages does not exist in the service. They do exist, we call them "latest". We actually want to use this especially when source bundle does not exist locally (i.e. in the case of microservices where source messages are not with the microservice that use it. e.g. VCF, Magna)

Third, let's not keep special logic for messages_source. The same fallback logic that we use for all locales apply to it—get from service if service is available, get from local if service is unavailable. To keep it simple for the users, we use the same "offlineResourcesBaseUrl" for local messages_source.json as well. Why keep the logic for "source" separate if it just means another type of bundle?

jessiejuachon: Regarding Huihui's comment (L2):

In debug mode, you should not see a log that says "can't find "source" locale". As I said, the code will iterate over the queue, and will not even reach "source" because it should find the fallback that it needs before reaching "source".  However, even if you see "can't find "xx" locale" in the log, that should be okay. It just means that the expected fallback locale does not exist, which can happen not just for "source" but for any other locale. It is not a bug.

If it does reach source, that is also okay. Imagine that in the future, we can actually enable a feature where the product team can provide a custom "source" data for L2. For now, "source" can be there with no harm.

src/main/java/com/vmware/vipclient/i18n/util/LocaleUtility.java
```diff
@@ -19,7 +19,8 @@
     
     // A locale fallback priority queue. For now, it only contains the default locale and the source messages (from "messages_source.json").
     private static List<Locale> fallbackLocales = new LinkedList<Locale>(Arrays.asList(getDefaultLocale(), Locale.forLanguageTag(ConstantsKeys.SOURCE)));
-    
+    private static List<Locale> l2FallbackLocales = new LinkedList<Locale>(Arrays.asList(getDefaultLocale(), Locale.ENGLISH));
```
jessiejuachon, line 22: I still disagree in having 2 separate fallback queue for L2 and L3. I think it creates unnecessary complexity of having to maintain 2 separate lists, 2 sets of setters, etc. Let's analyze the difference and similarities of L2 and L3 fallback behavior first. @gong-yu mentioned that we should have a discussion after 0.5.0 is released, so I created an issue here: https://github.com/vmware/singleton/issues/843

Questions:
1.  
Why does L2 need English in addition to the default locale that was set? Is it because the default locale may not be available while the English locale is always available? If yes, why? My concern is that we are tryin to remove any hardcoded locale from the code, but here you just hardcoded English.
2. What is the expected fallback behavior for L2? Isn't supposed to fallback to the default locale only, or should it also fallback to thew source locale (I do not see issue 758 anymore)? I  thought source locale is only meant for L3?

huihuiw01: We have discussed this item in last Friday's sprint review meeting, I have updated the conclusion to issue #758 , so I updated the code according to the conclusion, sorry for forgetting to let you know.

1. Yes, but I think we should consider correct behavior first, then code wise. Now the conclusion is the last fallback is English, so I have to hardcode it in code.

2. The conclusion of the correct fallback behavior for L2 is:
requested locale -> default locale -> English locale

jessiejuachon: Just talked to Demin, no need to hardcode English.

huihuiw01:
After discussion on Jan 22 2021, the final conclusion for L2 fallback is:
requested locale -> default locale

And L2 data for default locale should be always available, so no chance to fallback to source locale, so the case mentioned in this bug #806 is not valid, so bug #806 needn't be fixed and close this PR.

PR #807, opened 2020-09-25
G11n java client code clean up - remove logic of adding a cacheItem with an empty dataMap to cache for non-supported locales 

jessiejuachon:


jessiejuachon:
@lyiyu66 , tagging you just for your information. This is just the first step to solving 754: remove logic of adding a cacheItem with an empty dataMap to the cache for non-supported locales. We do not need this anymore because we are now getting the list of supported locales to check if a locale is not supported. Next step is to use the fallback locale immediately to avoid fetch from service.

jessiejuachon:
This is just the first step (code clean up) in solving 754. Here, I  have removed the logic of adding a cacheItem with an empty dataMap to the cache for non-supported locales. We do not need this anymore because we are now first getting the list of supported locales to check if a locale is supported or not. 

Next step is to use the fallback locale immediately to avoid fetch from service. This will be in another PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -102,19 +102,21 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
     			refreshCacheItemTask(cacheItem);
     		}
-    	} else { // Item is not in cache. Create and store cacheItem for the requested locale
+    	} else { // Item is not in cache.
+			ProductService ps = new ProductService(dto);
+			if (!ps.isSupportedLocale(Locale.forLanguageTag(dto.getLocale()))) { // Requested locale is not supported
+				Locale matchedLocale = LocaleUtility.pickupLocaleFromList(new LinkedList<>(ps.getSupportedLocales()), Locale.forLanguageTag(dto.getLocale()));
+				if (ps.isSupportedLocale(matchedLocale)) { // Requested locale matches a supported locale (eg. requested locale "fr_CA matches supported locale "fr")
+					MessagesDTO matchedLocaleDTO = new MessagesDTO(dto.getComponent(), matchedLocale.toLanguageTag(), dto.getProductID(), dto.getVersion());
+					return new ComponentService(matchedLocaleDTO).getMessages(null);
```
linr211, line 115: Seems recursive call is not good.

jessiejuachon: There is no recursive call in this line. Notice that there getMessages is invoked on a seaparate instance of ComponentService.

jessiejuachon: Also, this logic (lines 110-117) is not new. I just moved it out of the createCacheItem method that is called in line 120 into here for code clean up.

huihuiw01 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
linr211: I meant line 115. 'not good' means that it gets the messages twice: one for supported locale, another or not supported locale. The logic is duplicated, the locale should be handled before call getMessages.

Secondly, when i read 'return new ComponentService(matchedLocaleDTO).getMessages(null);', i don't get what it means by passing a null value.

Xiaochao8, line 115: It's obscure to pass null to getMessages. Hard to understand.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -102,19 +102,21 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+				}
+			}
+
+			//Create and store cacheItem for the requested locale
 			cacheItem = createCacheItem(fallbackLocalesIter);
```
Xiaochao8, line 120: Is it necessary to pass fallbackLocalesIter? I think createCacheItem shouldn't be affected by fallbackLocalesIter.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -102,19 +102,21 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+    		if (cacheItem.isExpired()) {
     			// Refresh the cacheItem in a separate thread
     			refreshCacheItemTask(cacheItem);
     		}
-    	} else { // Item is not in cache. Create and store cacheItem for the requested locale
+    	} else { // Item is not in cache.
+			ProductService ps = new ProductService(dto);
+			if (!ps.isSupportedLocale(Locale.forLanguageTag(dto.getLocale()))) { // Requested locale is not supported
+				Locale matchedLocale = LocaleUtility.pickupLocaleFromList(new LinkedList<>(ps.getSupportedLocales()), Locale.forLanguageTag(dto.getLocale()));
+				if (ps.isSupportedLocale(matchedLocale)) { // Requested locale matches a supported locale (eg. requested locale "fr_CA matches supported locale "fr")
```
Xiaochao8, line 113: Can't imagine getSupportedLocales is invoked three times in three code lines.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -131,69 +133,25 @@ private MessageCacheItem createCacheItem(Iterator<Locale> fallbackLocalesIter) {
...
-				return null;
-			}
-		};
-		FutureTask<MessageCacheItem> task = new FutureTask<>(callable);
-		Thread thread = new Thread(task);
-		thread.start();
-	}
-
 	private void refreshCacheItemTask(MessageCacheItem cacheItem) {
 		Callable<MessageCacheItem> callable = () -> {
```
Xiaochao8, line 149:
```suggestion
		Runnable runnable = () -> {
```

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: This code has been there before this PR, so I suggest we open another issue for your concern. becasue this PR is for code clean up only, not changing any logic.

The intention of this line is exactly what you mentioned. We want to get the matched locale **instead of** the actual requested locale if a match is found. This has been the same strategy for calling the fallback locale.. new ComponentService(fallbacklocaleDTO).getMessages() if needed. What is your suggestion on how to do that?

To understand the behavior when passing null to getMessages, read the javadoc of getMessages:
`/**
	 * Get MessageCacheItem from cache.
	 * The cache is refreshed if MessageCacheItem is expired or not found.
	 *
	 * @param fallbackLocalesIter The locale fallback queue to be used on failure. If null, there will be no fallback mechanism on failure so the message map will be empty.
	 *
	 * @return A MessageCacheItem whose data map is one of the following:
	 * <ul>
	 * 		<li>The messages in the requested locale</li>
	 * 	 	<li>The messages in a fallback locale</li>
	 * </ul>
	 */`

jessiejuachon: This has been there since before this PR, so I suggest you open another issue to address this specific concern. 

To understand the behavior when passing null to getMessages, read the javadoc of getMessages:

	 * @param fallbackLocalesIter The locale fallback queue to be used on failure. If null, there will be no fallback mechanism on failure so the message map will be empty.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
Xiaochao8: I think you can improve the LocaleUtility.pickupLocaleFromList to get enough information by only one call. Picking up once to get the supported locale or empty locale.

jessiejuachon: Ahhhm yes! I agree. I already have that lined up. However, as I said, this PR is only to clean up / remove any code related to adding an empty data set into the cache.

jessiejuachon:
@Xiaochao8 , @huihuiw01 , please focus on the intent of the PR. There are already 12 files modified which is quite a lot, so please refrain from discussing changes that are not related to the PR's intent (i.e. clean up code that is related to caching an empty data map for non-supported locales).

Xiaochao8 (approved):

huihuiw01 (approved):

PR #813, opened 2020-10-02
Fix for bug 812 G11n-java-client New workflow's Translation.getMessage must be able to use latest source messages from VIP service

jessiejuachon:
Fix for bug 812: G11n-java-client New workflow's Translation.getMessage must be able to fallback to "latest" messages from VIP service if messages_source.json is not present locally(offline) in the client

### DIFF
--- a/src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
+++ b/src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
@@ -55,8 +55,8 @@ public class ComponentService {
 		long timestampOld = cacheItem.getTimestamp();
 		DataSourceEnum dataSource = msgSourceQueueIter.next();
 		String localeOrig = dto.getLocale();
-		if (dataSource.equals(DataSourceEnum.VIP) && dto.getLocale().equals("source")) {
-			dto.setLocale("latest");
+		if (dataSource.equals(DataSourceEnum.VIP) && dto.getLocale().equals(ConstantsKeys.SOURCE)) {
+			dto.setLocale(ConstantsKeys.LATEST);
 		}
 		dataSource.createMessageOpt(dto).getComponentMessages(cacheItem);
 		long timestamp = cacheItem.getTimestamp();
@@ -67,7 +67,7 @@ public class ComponentService {
 
 		// If timestamp is 0, it means that cacheItem not yet in cache. So try the next data source.
 		// If locale is "source", use messages from DatasourceEnum.Bundle data store if it exists.
-		if (timestamp == 0 || dto.getLocale().equals("source")) {
+		if (timestamp == 0 || dto.getLocale().equals(ConstantsKeys.SOURCE)) {
 			// Try the next dataSource in the queue
 			refreshCacheItem(cacheItem, msgSourceQueueIter);
 		}
--- a/src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
+++ b/src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
@@ -54,18 +54,24 @@ public class ComponentService {
 
 		long timestampOld = cacheItem.getTimestamp();
 		DataSourceEnum dataSource = msgSourceQueueIter.next();
+		String localeOrig = dto.getLocale();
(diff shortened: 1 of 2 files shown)


src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -54,18 +54,24 @@ private void refreshCacheItem(final MessageCacheItem cacheItem, Iterator<DataSou
...
 			logger.debug(FormatUtils.format(ConstantsMsg.GET_MESSAGES_FAILED, dto.getComponent(), dto.getLocale(), dataSource.toString()));
 		}
+		dto.setLocale(localeOrig);
 
-		// If timestamp is not 0, it means that cacheItem is in the cache already.
-		// Otherwise, try the next dataSource in the queue.
-		if (timestamp == 0) {
+		// If timestamp is 0, it means that cacheItem not yet in cache. So try the next data source.
+		// If locale is "source", use messages from DatasourceEnum.Bundle data store if it exists.
+		if (timestamp == 0 || dto.getLocale().equals("source")) {
```
Xiaochao8, line 70: If getting latest feom service successfully, is it still necessary to get from local?

jessiejuachon: I am not so sure, actually. Is there a use case where a string has been updated only in local and not yet in remote? If yes, should the client use the local message or wait until message in service is updated?

Xiaochao8: If you think locale take priority over remote, then you should get from local first, then remote. Don't need to get from service first then get from local.

jessiejuachon: I have studied existing use cases. As a conclusion, let's keep it consistent with the behavior for all locales: Pull from remote when remote is available, from local if remote is not available. I'll remove the change from line 70.

Xiaochao8 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -54,18 +54,23 @@ private void refreshCacheItem(final MessageCacheItem cacheItem, Iterator<DataSou
 
 		long timestampOld = cacheItem.getTimestamp();
 		DataSourceEnum dataSource = msgSourceQueueIter.next();
+		String localeOrig = dto.getLocale();
+		if (dataSource.equals(DataSourceEnum.VIP) && dto.getLocale().equals(ConstantsKeys.SOURCE)) {
```
huihuiw01, line 58: Is it better unify the message file suffix in Singleton service and local bundle either to 'latest' or to 'source', then add the unified suffix to the fallback queue, in this case you needn't judge the data source and the requested locale here.

huihuiw01:
Looks like the change only resolve the problem for online and mix mode, do we need resolve the problem in offline mode which source bundles reside in other microservice's code?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: Yes, we will do that later on. We shall officially use "source" in the service (new version). Then in the client libraries, we have to support both versions of VIP service (old version with "latest" for backwards compatibility, and new version with "source").

In the meantime, for compatibility with current VIP service, we have to add this line.

jessiejuachon:
> Looks like the change only resolve the problem for online and mix mode, do we need resolve the problem in offline mode which source bundles reside in other microservice's code?

You are right, remote source bundles can only be supported if VIP service isn present. This is a known, by-design limitation of offline (only) mode support. Other limitations are: 

- Can not update localization at runtime (i.e. no localization bug fix unless you do a new product release/patch/hotfix)
- Can not add new language at runtime
- Can not share a single copy of localized bundle among services in a microservice environment. A copy of each resource bundle has to be packaged with each service

jessiejuachon:
Great observations, @huihuiw01 .

huihuiw01 (approved):

PR #816, opened 2020-10-12
Merge from 4.0.2 branch

Xiaochao8:


linr211 (approved):

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -96,67 +104,141 @@ public String getMessage(final Locale locale, final String component, final Stri
...
+        if (StringUtil.isEmpty(translation)) {
+            return "";
+        }
+
+        if (args != null) {
+            if (args instanceof Object[] && ((Object[]) args).length > 0) {
+                translation = FormatUtils.format(translation, locale, (Object[]) args);
+            } else if (args instanceof Map<?, ?> && ((Map) args).size() > 0) {
+                translation = FormatUtils.formatMsg(translation, locale, (Map<String, Object>) args);
+            }
```
huihuiw01, line 239: This code block appear multiple times in this class, so I think it can be wrapped as a method and put into FormatUtil class, then the code lines in this class can be reduced and code looks more clear.

huihuiw01 (changes requested):

linr211 (approved):

huihuiw01 (approved):

PR #823, opened 2020-10-13
G11n java client [BUG 754] Performance impact on the very first L3 fetch for a non-supported locale

jessiejuachon:


src/test/java/com/vmware/vip/i18n/LocaleTest.java
```diff
@@ -46,34 +42,72 @@ public void init() {
...
+        Assert.assertNull(LocaleUtility.pickupLocaleFromList(new HashSet<>(Arrays.asList(supportedLocales)), Locale.forLanguageTag("fil")));
+    }
+
+    /**
+     *  For any Chinese locale (zh-*) that is not supported,
+     *  return null so that fallback locale will be used even if "zh" is supported.
+     *  For any non-Chinese locale, return the best match (e.g. 'de' if 'de-DE' is not supported).
+     */
+    @Test
+    public void testPickupLocaleFromList_special_case_zh_HK() {
```
Xiaochao8, line 96: Add one more test case: zh is supported, zh-Hans and zh-Hant aren't supported.  Request zh-Hant will return en-US.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -108,38 +106,48 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+					cacheItem = getMessagesFallbackLocale(fallbackLocalesIter);
+			} else { // Requested locale is not supported
+				Locale matchedLocale = LocaleUtility.pickupLocaleFromList(ps.getSupportedLocales(), locale);
+				if (matchedLocale != null) { // Requested locale matches a supported locale (eg. requested locale "fr_CA matches supported locale "fr")
 					MessagesDTO matchedLocaleDTO = new MessagesDTO(dto.getComponent(), matchedLocale.toLanguageTag(), dto.getProductID(), dto.getVersion());
-					return new ComponentService(matchedLocaleDTO).getMessages();
-				}
+					cacheItem = new ComponentService(matchedLocaleDTO).getMessages();
+				} else  // Requested locale is not supported and does not match any supported locales
+					cacheItem = getMessagesFallbackLocale(fallbackLocalesIter);
```
huihuiw01, line 120: I think if you handle locale before looking for the messages from cache, the logic here may be  simplifier. 

In my opinion, the logic should be as below:
1. Handle locale:
Locale matchedLocale = LocaleUtility.pickupLocaleFromList(ps.getSupportedLocales(), locale);
2. If matchedLocale is not empty, use matchedLocale to fetch messages from data source;
if matchedLocale is empty, directly use fallback locale to fetch messages from data source;
3. if above fetch failed, fallback;

src/main/java/com/vmware/vipclient/i18n/util/LocaleUtility.java
```diff
@@ -147,43 +142,15 @@ public static boolean isSameLocale(String locale1, String locale2) {
...
-        // With Chinese locale which is not configured/supported in web.xml, it
-        // will return 'en_US' as default to meet the usage custom of Chinese,
-        // e.g. for 'zh-HK' from client(browser) which is not
-        // configured/supported yet, it will return 'en_US';
-        // Other locale, like 'de-DE' 'ja-JP' etc.,
-        // it will return 'de' 'ja'(main/parent language).
-        if (langLocale != null
-                && (!langLocale.getLanguage().equalsIgnoreCase("zh"))) {
-            return new Locale(langLocale.getLanguage());
+        // For any Chinese locale (zh-*) that is not supported, use the fallback locale even if "zh" is supported.
```
huihuiw01, line 149: Why use fallback locale instead of 'zh'?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -108,38 +106,48 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
     	}
     	return cacheItem;
     }
 
+    /**
+     * Gets the messages in the next fallback locale by passing the next fallback locale DTO to a new instance of ComponentService
+	 * and then invoking {@link #getMessages(Iterator)}.
+	 * @param fallbackLocalesIter The fallback locale queue to use in case of failure. If null, no locale fallback will be applied.
+     */
+    private MessageCacheItem getMessagesFallbackLocale(Iterator<Locale> fallbackLocalesIter) {
```
Xiaochao8, line 131: fallbackLocalesIter may not be a parameter because it can be got by LocaleUtility.getFallbackLocales().iterator()

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
Xiaochao8, line 120: We can unify this these 2 situations. If matchedLocale == null, then set locale to fallback locale. Get its translation afterwards.

src/main/java/com/vmware/vipclient/i18n/util/LocaleUtility.java
jessiejuachon: It has been like that previously. I am guessing it was an old requirement.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: I understand what you are saying. The problem with this is due to the locale being part of the DTO. If the matched locale is not null, there are 2 possible scenarios:

a. The requested locale is a supported locale (exact match) -- lines 110-113
b. The requested locale is not an exact match (e.g. fr-CA which is not supported matches fr which is supported) -- lines 117-118

If it is (a), createCacheItem in line 111 fetches the matched locale and caches it. Take note that it uses the **dto's locale in the cache key**.
If it is (b), createCacheItem must be done on the supported locale (fr), not the requested locale (fr-CA). If I directly call createCacheItem for this case, the cache key will have fr-CA for the same set of fr messages. causing duplicate sets of messages in the cache, which is not good.

jessiejuachon: You have a similar comment as Huihui. Pasting my reply below:

The problem with this is due to the legacy code where locale is part of the DTO. If the matched locale is not null, there are 2 possible scenarios:

a. The requested locale is a supported locale (exact match) -- lines 110-113
Here, createCacheItem in line 111 fetches the matched locale and caches it. Take note that it uses the dto's locale to generate the cache key.
b. The requested locale is not an exact match (e.g. fr-CA which is not supported matches fr which is supported) -- lines 117-118
Here, createCacheItem must be done on the supported locale (fr), not the requested locale (fr-CA). If I directly call createCacheItem for this case, the cache key will have fr-CA for the same set of fr messages, causing duplicate sets of messages in the cache, which is not good. So, call ComponentService(matchedLocaleDTO).getMessages instead.

This is why these 2 scenarios are handled differently. We can refactor this in the future and extract locale out of the DTO, but I did not want to touch that legacy code for now.

Xiaochao8: I think that our comments are different. I'm saying 115~120.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -108,38 +105,48 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
     	return cacheItem;
     }
 
+    /**
+     * Gets the messages in the next fallback locale by passing the next fallback locale DTO to a new instance of ComponentService
+	 * and then invoking {@link #getMessages(Iterator)}.
+	 * @param fallbackLocalesIter The fallback locale queue to use in case of failure. If null, no locale fallback will be applied.
+     */
+    private MessageCacheItem getFallbackLocaleMessages(Iterator<Locale> fallbackLocalesIter) {
+		if (!dto.getLocale().equals(ConstantsKeys.SOURCE) && fallbackLocalesIter != null && fallbackLocalesIter.hasNext()) {
```
Xiaochao8, line 131: Is it necessary to judge dto locale is same as the fallback locale?

src/main/java/com/vmware/vipclient/i18n/util/LocaleUtility.java
```diff
@@ -146,9 +146,12 @@ public static Locale pickupLocaleFromList(Set<Locale> locales,
...
 
-        // For any Chinese locale (zh-*) that is not supported, use the fallback locale even if "zh" is supported.
-        if (bestMatch != null && bestMatch.toLanguageTag().equals("zh") && !preferredLocale.toLanguageTag().equals("zh")) {
-            return null;
+        // For any Chinese locale (zh-*) that is not supported (except for zh-Hans and zh-Hant), use the fallback locale even if "zh" is supported.
+        if (preferredLocale.getLanguage().equals("zh")) {
+            Locale zhLocale = fmtToMappedLocale(preferredLocale);
+            if (!locales.contains(zhLocale) && !zhLocale.toLanguageTag().equals("zh-Hans") && !zhLocale.toLanguageTag().equals("zh-Hant")) {
+                return null;
+            }
```
Xiaochao8:
```suggestion
       if (preferredLocale.getLanguage().equals("zh")) {
            Locale zhLocale = fmtToMappedLocale(preferredLocale);
            if (!locales.contains(zhLocale)) {
                String langTag = zhLocale.toLanguageTag();
                if (langTag.equals("zh-Hans") || langTag.equals("zh-Hant")) {
                    return null;
                }
            }
        }
```

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: I do not understand then. Could you please provide a code suggestion?

Xiaochao8: Please check this code suggestion for 115 - 120:

ProductService ps = new ProductService(dto);
			Locale localeToQuery = Locale.forLanguageTag(dto.getLocale());
			if (!ps.isSupportedLocale(localeToQuery)) {
				localeToQuery = LocaleUtility.pickupLocaleFromList(ps.getSupportedLocales(), localeToQuery);
				if (localeToQuery == null) {// Requested locale is not supported and does not match any supported locales
					if (!dto.getLocale().equals(ConstantsKeys.SOURCE) && fallbackLocalesIter != null && fallbackLocalesIter.hasNext()) {
						localeToQuery = fallbackLocalesIter.next();
					}
				} else {
					return new MessageCacheItem();
				}
			}

			MessagesDTO matchedLocaleDTO = new MessagesDTO(dto.getComponent(), localeToQuery.toLanguageTag(), dto.getProductID(), dto.getVersion());
			cacheItem = new ComponentService(matchedLocaleDTO).getMessages();
    	}

jessiejuachon: Thanks for the code suggestion. It will work, but I think what I currently have is cleaner and has less lines of code, don't you think? Basically, your `if (!dto.getLocale().equals(ConstantsKeys.SOURCE) && fallbackLocalesIter != null && fallbackLocalesIter.hasNext()) `

is a repeat of line 132.

src/main/java/com/vmware/vipclient/i18n/util/LocaleUtility.java
jessiejuachon: I understand you want it to be more readable. There is incorrect logic in your suggestion though. It should be `if (!langTag.equals("zh-Hans") && !langTag.equals("zh-Hant"))` in line 160. Please review the new code.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: It's not necessary, but it is an improvement. If we do not check for dto.getLocale == fallback and it happens, it will just try to fetch for the dto locale again and will fail again, and move on to the next item in the queue. Nonetheless, I have added the check for optimization.

Moreover, I hsve removed the following restriction to allow falling back to SOURCE messages:
`!dto.getLocale().equals(ConstantsKeys.SOURCE)`

src/main/java/com/vmware/vipclient/i18n/util/LocaleUtility.java
```diff
@@ -147,43 +142,21 @@ public static boolean isSameLocale(String locale1, String locale2) {
...
-                if (((preferredScript.equalsIgnoreCase("")) && (configuredScript
-                        .equalsIgnoreCase("")))
-                        || ((!preferredScript.equalsIgnoreCase("")) && (preferredScript
-                                .equalsIgnoreCase(configuredScript)))) {
-                    return configuredLocale;
-                }
-                langLocale = langLocale == null ? configuredLocale : langLocale;
+    public static Locale pickupLocaleFromList(Set<Locale> locales,
+                                              Locale preferredLocale) {
+        Locale bestMatch = Locale.lookup(Arrays.asList(new Locale.LanguageRange(fmtToMappedLocale(preferredLocale).toLanguageTag())), locales);
```
Xiaochao8, line 147:
```suggestion
		Locale localeObject = fmtToMappedLocale(preferredLocale);
		Locale bestMatch = Locale.lookup(Arrays.asList(new Locale.LanguageRange(localeObject.toLanguageTag())),
				locales);
```

src/main/java/com/vmware/vipclient/i18n/util/LocaleUtility.java
```diff
@@ -147,43 +142,21 @@ public static boolean isSameLocale(String locale1, String locale2) {
...
+
+        // For any Chinese locale (zh-*) that is not supported (except for zh-Hans and zh-Hant), use the fallback locale even if "zh" is supported.
+        if (preferredLocale.getLanguage().equals("zh")) {
+            Locale zhLocale = fmtToMappedLocale(preferredLocale);
+            if (!locales.contains(zhLocale)) {
+                String langTag = zhLocale.toLanguageTag();
+                if (!langTag.equals("zh-Hans") && !langTag.equals("zh-Hant"))
+                    return null;
```
Xiaochao8, line 155:
```suggestion
		// With Chinese locale which is not configured/supported in web.xml, it
		// will return 'en_US' as default to meet the usage custom of Chinese,
		// e.g. for 'zh-HK' from client(browser) which is not
		// configured/supported yet, it will return 'en_US';
		// Other locale, like 'de-DE' 'ja-JP' etc.,
		// it will return 'de' 'ja'(main/parent language).
		if (bestMatch != null && bestMatch.getLanguage().equals("zh")) {
			if (!locales.contains(localeObject)) {
				return null;
			}
		}
```

jessiejuachon: @Xiaochao8 , this change fails for test LocaleTest. testPickupLocaleFromListZh

Xiaochao8 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
huihuiw01: Can reseting dto's locale after locale matching resolve the problem you mentioned?

I think when matchedLocale is not empty, all later logic(no matter find from cache or find from data source) should use matchedLocale, request locale is useless.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -113,38 +110,52 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+     */
+    private MessageCacheItem getFallbackLocaleMessages(Iterator<Locale> fallbackLocalesIter) {
+		if (fallbackLocalesIter != null && fallbackLocalesIter.hasNext()) {
+			Locale fallbackLocale = fallbackLocalesIter.next();
+			if (fallbackLocale.toLanguageTag().equals(dto.getLocale())) {
+				return getFallbackLocaleMessages(fallbackLocalesIter);
+			}
+			// Use MessageCacheItem of the next fallback locale.
+			MessagesDTO fallbackLocaleDTO = new MessagesDTO(dto.getComponent(), fallbackLocale.toLanguageTag(), dto.getProductID(), dto.getVersion());
+			return new ComponentService(fallbackLocaleDTO).getMessages(fallbackLocalesIter);
```
huihuiw01, line 143: In this way when dto locale is 'source' and source messages not found, it will fallback to default locale.

huihuiw01 (changes requested):

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: If dto locale is source, it means it has reached the end of the fallback queue. Hence, it won't enter the if in line 136 because  `fallbackLocalesIter.hasNext()` is false.

huihuiw01: I mean when request locale is 'source', which make dto locale is 'source'

jessiejuachon: Do you mean the preferred locale from the user is "source"? What use case do you have in mind where locale will be "source"? I suppose "source" will only be used as fallback.

Nonetheless, if user actually sends "source" as the preferred locale and source messages are not found, then yes, it should go though the usual locale fallback queue. What is your concern with this?

jessiejuachon: Discussed this offline. Huihui is referring to this issue which is not within the scope of this PR: https://github.com/vmware/singleton/issues/839

huihuiw01 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -103,48 +100,62 @@ public MessageCacheItem getMessages() {
...
-			}
+			Locale locale = Locale.forLanguageTag(dto.getLocale());
+			if (ps.isSupportedLocale(locale) || VIPCfg.getInstance().isPseudo()) {
+				cacheItem = createCacheItem(); // Fetch for the requested locale from data store, create cacheItem and store in cache
+				if (cacheItem.getCachedData().isEmpty())  // Failed to fetch messages for the requested locale
+					cacheItem = getFallbackLocaleMessages(fallbackLocalesIter);
+			} else   // Requested locale is not supported and does not match any supported locales
+				cacheItem = getFallbackLocaleMessages(fallbackLocalesIter);
+		}
+		return cacheItem;
```
Xiaochao8, line 121: Suggest code for this method, then doLocaleMatching can be removed:

		Locale matchedLocale = LocaleUtility.pickupLocaleFromList(new ProductService(dto).getSupportedLocales(),
				Locale.forLanguageTag(dto.getLocale()));
		if (matchedLocale != null) {
			dto.setLocale(matchedLocale.toLanguageTag());
			cacheItem = new CacheService(dto).getCacheOfComponent();
			if (cacheItem != null) { // Item is in cache
				if (cacheItem.isExpired())
					refreshCacheItemTask(cacheItem); // Refresh the cacheItem in a separate thread
			} else { // Item is not in cache.
				cacheItem = createCacheItem(); // Fetch for the requested locale from data store, create cacheItem and store in cache
			}
		}
		
		if (cacheItem == null || cacheItem.getCachedData().isEmpty()) // Failed to fetch messages for the requested locale
			cacheItem = getFallbackLocaleMessages(fallbackLocalesIter);

		return cacheItem;

Xiaochao8 (approved):

Issue #824, opened 2020-10-13
[ENHANCEMENT] Code clean up in the java client library - Sprint 90


jessiejuachon:
**Is your feature request related to a problem? Please describe.**
This is a placeholder issue for anything related to code clean up of the g11n-java-client in Sprint 90.

**Describe the solution you'd like**
One or more PRs may be submitted. Each PR must not change or add any logic or functionality, and must be solely for cleaning up code (e.g. removing unused code, marking API methods or classes as deprecated, renaming methods or classes for better readability)

jessiejuachon:
Ready for testing (regression test only, no new features)

PR #829, opened 2020-10-19
G11n java client - code cleanup - removing null checks for properties/variables that will never be null

jessiejuachon:


src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
```diff
@@ -73,18 +73,8 @@ public synchronized boolean isExpired() {
     	if (VIPCfg.getInstance().getCacheExpiredTime() != 0) {
     		return false;
     	}
-    	
-    	Long responseTimeStamp = this.getTimestamp();
-    	if (responseTimeStamp == null) {
-    		return true;
-    	}
-    	
-    	Long maxAgeResponse = this.getMaxAgeMillis();
```
Xiaochao8, line 82: I don't agree. By setCacheItem, it can be null.

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
```diff
@@ -73,18 +73,8 @@ public synchronized boolean isExpired() {
...
-    	if (maxAgeResponse != null) {
-    		maxAgeMillis = maxAgeResponse;
-    	}
-    		  	
-    	return System.currentTimeMillis() - responseTimeStamp > maxAgeMillis;
+
+    	return System.currentTimeMillis() - this.getTimestamp() > this.getMaxAgeMillis();
```
huihuiw01, line 77: Same comment with xiaochao's comment to my PR:
1. change '>' to '>=';
2. add this line to synchronized block, or the timestamp and the maxAgeMillis may not match.

huihuiw01 (changes requested):

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
jessiejuachon: Hi Huihui! The whole method is already synchronized.
I will change the '>' to '>=' in another PR because these quick clean up PRs must have no code logic change in them. https://github.com/vmware/singleton/issues/824

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
jessiejuachon: You are correct! I already have listed it as todo: _Code clean up -  remove MessageCacheItem.setCacheItem (MessageCacheItem cacheItem)_

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
jessiejuachon: 1.  Yes, I have that listed for another PR. It's not related to this.
2.  The while method is already synchronized. Nonetheless, it is also not related to the PR.

src/main/java/com/vmware/vipclient/i18n/base/cache/MessageCacheItem.java
jessiejuachon: Do you mean the timestamp property? Its type is long, not Long. The default value of long is 0. It cannot be null because it is not a Java object.

Xiaochao8: I mean maxAgeMillis. It's Long type.

jessiejuachon: How can it be null? There are 3 setCacheItem methods: lines 33, 38 and 46. All of them eventually execute lines 42-43:
`if (maxAgeMillis != null)
			this.maxAgeMillis = maxAgeMillis;`

Xiaochao8 (approved):

huihuiw01 (approved):

PR #830, opened 2020-10-19
G11n java client - vipcfg cleanup

jessiejuachon:
Code cleanup - use setter methods; remove unused code

### DIFF
--- a/src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
+++ b/src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
@@ -108,12 +108,7 @@ public class VIPCfg {
             cfg.isSubInstance = true;
             VIPCfg.moduleCfgs.put(productName, cfg);
         }
-
-        if (VIPCfg.moduleCfgs.containsKey(productName) && VIPCfg.moduleCfgs.get(productName) != null) {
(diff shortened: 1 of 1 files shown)


huihuiw01 (dismissed):

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -319,6 +312,7 @@ public String getVipServer() {
 
     public void setVipServer(String vipServer) {
         this.vipServer = vipServer;
+        this.addMsgOriginsQueue(DataSourceEnum.VIP);
```
Xiaochao8, line 315: Because service takes priority over offline bundles, so we should make sure service will be before offline bundles.

jessiejuachon: We make sure of this sequence (VIP first before offline) when VIPCfg.initialize(configfile) is used. However, if product team chooses to change the priority, they should be able to do so by calling the individual setters in the order they prefer.

Xiaochao8 (approved):

Xiaochao8 (approved):

huihuiw01 (approved):

PR #848, opened 2020-10-29
Code clean up in VIPCfg. Remove TranslationCacheManager member from VIPCfg because TranslationCacheManager is using Singleton design pattern.

jessiejuachon:


Xiaochao8 (dismissed):

src/main/java/com/vmware/vipclient/i18n/base/cache/TranslationCacheManager.java
```diff
@@ -28,7 +28,7 @@
     private TranslationCacheManager() {
     }
 
-    public static synchronized TranslationCacheManager createTranslationCacheManager() {
+    public static synchronized TranslationCacheManager getTranslationCacheManager() {
```
huihuiw01, line 31: Is it better rename the method name as 'getInstance'?

jessiejuachon: Made this name change. Please approve, @Xiaochao8 and @huihuiw01 .Thanks!

Xiaochao8 (dismissed):

src/main/java/com/vmware/vipclient/i18n/base/L10nTypeEnum.java
```diff
@@ -0,0 +1,33 @@
+package com.vmware.vipclient.i18n.base;
+
+import com.vmware.vipclient.i18n.VIPCfg;
+import com.vmware.vipclient.i18n.base.cache.FormattingCache;
+import com.vmware.vipclient.i18n.base.cache.MessageCache;
+
+public enum L10nTypeEnum {
```
huihuiw01, line 7: Looks like this new added class is not used, so what's the purpose of adding it?

jessiejuachon: Added this by mistake. Removing it now.

huihuiw01 (approved):

Xiaochao8 (approved):

PR #855, opened 2020-11-03
Fix for 839: Do not use locale fallback when getting source locale messages and collectedSourceMsg in TranlationMessage.getMessage

jessiejuachon:


jessiejuachon:
Note: I had to remove some unnecessary pseudo-related logic from ComponentBasedOpt (even if they seem to be unrelated to this PR). This is because pseudo-related tests were failing after the changes in TranslationMessage. The changes in ComponentBasedOpt actually makes it cleaner and fixes the tests.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
```diff
@@ -35,9 +35,6 @@ public ComponentBasedOpt(final MessagesDTO dto) {
     public void getComponentMessages(MessageCacheItem cacheItem) {
         String url = V2URL.getComponentTranslationURL(this.dto,
                 VIPCfg.getInstance().getVipService().getHttpRequester().getBaseURL());
-        if (ConstantsKeys.LATEST.equals(this.dto.getLocale())) {
-            url = url.replace("pseudo=false", "pseudo=true");
-        }
```
huihuiw01, line 40: Could you explain what problem do these lines resolve before you remove them?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
```diff
@@ -81,10 +78,6 @@ private JSONObject getMsgsJson(Map<String, Object> response) {
 			if (null == responseStr || responseStr.equals(""))
 				return null;
 			else {
-				if (ConstantsKeys.LATEST.equals(this.dto.getLocale())) {
-					responseStr = responseStr.replace(ConstantsKeys.PSEUDOCHAR, "");
-				}
-
```
huihuiw01, line 87: Could you explain what problem do these lines resolve before you remove them?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
jessiejuachon: Nothing, really. Here, it says that if locale is latest", force it to use pseudo=true, which means the result will have #@ around it. This does not even makes sense because if you look at lines 84-87 below, it was stripping all the #@ from the response anyway, forcing the messages to ALWAYS look like pseudo=false if locale is "latest". 

We do not need to force pseudo=false for "latest". Removing this allows the product team to apply the same pseudo configuration, whether it is true or false, to all locale bundles including "latest".

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/server/ComponentBasedOpt.java
jessiejuachon: Same comment as above.

Xiaochao8 (approved):

huihuiw01 (approved):

PR #858, opened 2020-11-03
Code clean up - Do not use the singleton design pattern on VIPCfg

jessiejuachon:
VIPCfg.java was originally using the Singleton design pattern (single static VIPCfg object instance for the entire application). However, the use case of having multiple VIP products in the same app was introduced (a.k.a subInstance). The singleton design pattern is not applicable in this anymore as multiple instances of VIPCfg have to be created and stored.

Solution:
1. Remove the static "gcInstance" from VIPCfg. Store it in VIPConfigs.java.
2. Remove moduleCfgs from VIPCfg. Store it in VIPConfigs.java.

### DIFF
--- a/src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
+++ b/src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
@@ -35,8 +35,6 @@ public class VIPCfg {
     Logger                             logger        = LoggerFactory.getLogger(VIPCfg.class);
 
     // define global instance
-    private static VIPCfg              gcInstance;
-    private static Map<String, VIPCfg> moduleCfgs    = new HashMap<String, VIPCfg>();
     private VIPService                 vipService;
     private TranslationCacheManager    translationCacheManager;
 
@@ -79,34 +77,28 @@ public class VIPCfg {
 
     private boolean isSubInstance = false;
 
-    private VIPCfg() {
+    public VIPCfg() {
 
     }
 
     /**
-     * create a default instance of VIPCfg
-     * 
-     * @return
+     * @deprecated Use the {@link com.vmware.vipclient.i18n.VIPConfigs#getMainCfg() getMainCfg} method.
      */
-    public static synchronized VIPCfg getInstance() {
-        if (gcInstance == null) {
-            gcInstance = new VIPCfg();
-        }
-        return gcInstance;
+    public static VIPCfg getInstance() {
+        return VIPConfigs.getMainCfgInstance();
     }
 
     /**
-     * create a default instance of VIPCfg
-     *
-     * @return
+     * @deprecated Use the {@link com.vmware.vipclient.i18n.VIPConfigs#getCfg(String) getCfg} method.
      */
     public static synchronized VIPCfg getSubInstance(String productName) {
-        if (!VIPCfg.moduleCfgs.containsKey(productName)) {
+        if (!VIPConfigs.contains(productName)) {
             VIPCfg cfg = new VIPCfg();
             cfg.isSubInstance = true;
-            VIPCfg.moduleCfgs.put(productName, cfg);
+            cfg.setProductName(productName);
+            VIPConfigs.addCfg(cfg);
         }
-        return VIPCfg.moduleCfgs.get(productName);
+        return VIPConfigs.getCfg(productName);
     }
 
     /**
@@ -135,7 +127,7 @@ public class VIPCfg {
 
         if (prop.containsKey("productName"))
             this.setProductName(prop.getString("productName"));
-        if (this.isSubInstance() && !VIPCfg.moduleCfgs.containsKey(this.productName)) {
+        if (this.isSubInstance() && !VIPConfigs.contains(this.productName)) {
             throw new VIPClientInitException(
                     "Can't not initialize sub VIPCfg instance, the product name is not defined in config file.");
         }
--- a/src/main/java/com/vmware/vipclient/i18n/VIPConfigs.java
+++ b/src/main/java/com/vmware/vipclient/i18n/VIPConfigs.java
@@ -1,0 +1,33 @@
+package com.vmware.vipclient.i18n;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class VIPConfigs {
+    private static final Map<String, VIPCfg> configs = new HashMap<>();
+    private static VIPCfg mainCfg;
+
+    public static VIPCfg getMainCfg() {
+        return mainCfg;
+    }
+
+    public static synchronized VIPCfg getMainCfgInstance() {
+        if (getMainCfg() == null) {
+            mainCfg = new VIPCfg();
+        }
+        return VIPConfigs.getMainCfg();
+    }
+
+    public static VIPCfg getCfg(String productName) {
+        return configs.get(productName);
+    }
+
+    public static void addCfg(VIPCfg cfg) {
+        configs.put(cfg.getProductName(), cfg);
+    }
+
+    public static boolean contains(String productName) {
+        return configs.containsKey(productName);
+    }
+
+}


src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -30,13 +28,11 @@
 /**
  * a class uses to define the global environment setting for I18nFactory
  */
-public class VIPCfg {
+public class VIPCfg extends VIPConfigs {
```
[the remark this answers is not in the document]
jessiejuachon: The protected addCfg method.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -79,34 +75,22 @@ public void setSubInstance(boolean subInstance) {
 
     private boolean isSubInstance = false;
 
-    private VIPCfg() {
+    public VIPCfg() {
```
Xiaochao8, line 78: Why public this method? Is it necessary?

src/main/java/com/vmware/vipclient/i18n/VIPConfigs.java
```diff
@@ -0,0 +1,38 @@
...
+public class VIPConfigs {
+    private static final Map<String, VIPCfg> configs = new HashMap<>();
+    private static VIPCfg mainCfg;
+
+    public static synchronized VIPCfg getMainCfg() {
+        if (mainCfg == null) {
+            mainCfg = new VIPCfg();
+        }
+        return mainCfg;
+    }
```
Xiaochao8, line 19:
```suggestion
    public static VIPCfg getMainCfg() {
        if (mainCfg == null) {
            synchronized (VIPConfigs.class) {
                  if (mainCfg == null) {
                      mainCfg = new VIPCfg();
                  }
            }
        }
        return mainCfg;
    }
```

src/main/java/com/vmware/vipclient/i18n/VIPConfigs.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2019-2020 VMware, Inc.
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package com.vmware.vipclient.i18n;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class VIPConfigs {
```
Xiaochao8, line 10: Why introduct a new class? Is it necessary?

src/main/java/com/vmware/vipclient/i18n/VIPConfigs.java
```diff
@@ -0,0 +1,38 @@
...
+            cfg.setProductName(productName);
+        }
+        return configs.get(productName);
+    }
+
+    protected static void addCfg(VIPCfg cfg) {
+        configs.put(cfg.getProductName(), cfg);
+    }
+
+    public static boolean contains(String productName) {
```
Xiaochao8, line 34: make this synchronized ?

src/main/java/com/vmware/vipclient/i18n/VIPConfigs.java
```diff
@@ -0,0 +1,38 @@
...
+    private static VIPCfg mainCfg;
+
+    public static synchronized VIPCfg getMainCfg() {
+        if (mainCfg == null) {
+            mainCfg = new VIPCfg();
+        }
+        return mainCfg;
+    }
+
+    public static VIPCfg getCfg(String productName) {
```
Xiaochao8, line 21: make this synchronized ?

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -68,44 +64,19 @@
...
+        return VIPCfgFactory.getCfg();
     }
 
     /**
-     * create a default instance of VIPCfg
-     *
-     * @return
+     * @deprecated Use the {@link VIPCfgFactory#getCfg(String) getCfg} method.
      */
     public static synchronized VIPCfg getSubInstance(String productName) {
```
Xiaochao8, line 77: synchronized is redundant?

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * A lazy-loaded singleton Map<String, VIPCfg> configs using the "initialization-on-demand holder" design pattern
+     */
+    private static class ConfigsHolder {
+        static final Map<String, VIPCfg> configs = new HashMap<>();
+    }
+
+    public static VIPCfg getCfg(String productName) {
+        if (!contains(productName)) {
+            synchronized (ConfigsHolder.configs) {
```
Xiaochao8, line 33: synchronized should be applied in the method. If not, line40 will read the configs without locking.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
```diff
@@ -0,0 +1,51 @@
...
+        return ConfigsHolder.configs.get(productName);
+    }
+
+    protected static void addCfg(VIPCfg cfg) {
+        ConfigsHolder.configs.put(cfg.getProductName(), cfg);
+    }
+
+    private static boolean contains(String productName) {
+        return ConfigsHolder.configs.containsKey(productName);
+    }
```
Xiaochao8, line 49: Access the ConfigsHolder.configs should be synchronized and in only one location in lines from 32-40.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: This is deprecated anyway. Do we still want to change this?

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: That's exactly what we want, for line 40 to read without locking. If the productName is already there, the thread should not have to acquire a lock. Otherwise, performance will be affected.

The lock should only be acquired if writing to ConfigsHolder.configs. Actually, now that I am, thinking about it, do we even need to lock here? What should happen if product application's code tries to add the same product name twice? Should we throw an exception, or just allow to overwrite the old entry?

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: Why does reading from ConfigsHolder.configs have to be synchronized? What is the thread collision use case you are trying to prevent here?

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
Xiaochao8: It's better to change because the lock has been moved to other places.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
Xiaochao8: If other threads are writing when reading, is this a problem?

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
Xiaochao8: If other threads are writing when reading, is this a problem?

jessiejuachon: It's not a problem. Any thread will read the current state, whether it is before, during or after a synchronized thread finishes writing. That should be okay. 

The lock will only be acquired once, when adding a VIPCfg instance to ConfigsHolder.configs **for the first time**. After that, any thread can get the same VIPCfg instance and read/write it's properties.

Think of it this way: this is similar to Singleton design pattern too, except that the VIPCfg instance is **per product**, not the entire application. In singleton design pattern, only the creation of the instance needs to be synchronized, not the reading of the instance, correct?

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: First, contains() method is private and only used inside getCfg(), so I think this is the same conversation as above.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: sure

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
Xiaochao8: Please refer to this:
(<a href='https://stackoverflow.com/questions/18542037/how-to-prove-that-hashmap-in-java-is-not-thread-safe'>multithreading - How to prove that HashMap in java is not thread-safe - Stack Overflow</a>)

jessiejuachon: Actually, please pause the discussion on this. I am making some changes to the solution, and the hashmap may not be needed anymore.

jessiejuachon: Are you referring to the reply regarding temporary empty array while resizing that may return null? Point well taken. Thanks for pointing that out. 

 I will also make some more changes aside from this. Will request another review when ready. Thanks!

jessiejuachon: I just changed the HashMap to ConcurrentHashMap.  This will take care of the resizing issue.
The unsynchronized call in line 32 `if (!contains(productName)) {` takes care of the fast read when VIpCfg exists.
The synchronized call in 34 `if (!contains(productName)) {` makes sure that the first thread to enter the synchronized block gets to instantiate VIPCfg.

jessiejuachon: * I just changed the HashMap to ConcurrentHashMap. This will take care of the resizing issue.
* The unsynchronized call in line 32 `if (!contains(productName)) {` takes care of the fast read when VIPCfg already exists.
* The synchronized call in line 34 `if (!contains(productName)) {` makes sure that the first thread to enter the synchronized block gets to instantiate VIPCfg.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
```diff
@@ -0,0 +1,50 @@
...
+                if (!contains(productName)) {
+                    VIPCfg cfg = new VIPCfg();
+                    cfg.setProductName(productName);
+                }
+            }
+        }
+        return ConfigsHolder.configs.get(productName);
+    }
+
+    static void addCfg(VIPCfg cfg) {
```
Xiaochao8, line 42: Make this private and call this in line 35. In this way, only one method takes charge of writing config map. This is clear.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
```diff
@@ -68,44 +64,22 @@
...
-    }
-    
-    public void setSubInstance(boolean subInstance) {
-        isSubInstance = subInstance;
-    }
-
-    private boolean isSubInstance = false;
-
-    private VIPCfg() {
+    VIPCfg() {
```
Xiaochao8, line 67: This is unnecessary.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class VIPCfgFactory {
+
+    /**
+     * A lazy-loaded singleton VIPCfg using the "initialization-on-demand holder" design pattern
+     */
+    private static class GlobalCfgHolder {
+        static final VIPCfg globalCfg = new VIPCfg();
```
Xiaochao8, line 16: Suggest to save main product's name here, and save the real instance into the 'configs' map. The initial name can be "default".

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static VIPCfg getGlobalCfg() {
+        return GlobalCfgHolder.globalCfg;
+    }
+
+    /**
+     * A lazy-loaded singleton Map<String, VIPCfg> configs using the "initialization-on-demand holder" design pattern
+     */
+    private static class ConfigsHolder {
+        static final Map<String, VIPCfg> configs = new ConcurrentHashMap<>();
```
Xiaochao8, line 27: Because config name may be changed separately, there should be a method to deal with this to change the key.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: I have deprecated the capability to instantiate with no product name. See new code changes.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: I have deprecated the capability to instantiate VIPCfg with no product name. See new code changes.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
Xiaochao8: No. I don't think there are product using default constructor because VIPCfg is a singleton before.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+        return ConfigsHolder.configs.get(productName);
+    }
+
+    /**
+     * @deprecated Use {@link #getCfg(String) getCfg} instead.
+     * This method was added for backwards compatibility with deprecated {@link VIPCfg#getInstance()} and {@link VIPCfg#setProductName(String)} methods.
+     */
+    static void addCfg(String productName, VIPCfg cfg) {
```
Xiaochao8, line 58: You can remove this directly because it's new.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public static VIPCfgWrapper getGlobalCfg() {
+        return GlobalCfgHolder.globalCfg;
+    }
+
+    /**
+     * @deprecated This method was added for backwards compatibility with deprecated {@link VIPCfg#getInstance() getInstance}.
+     * Use the {@link #getGlobalCfg() getGlobalCfg} method.
+     */
+    public static VIPCfgWrapper getCfg() {
```
Xiaochao8, line 28: You can remove this directly because it's new.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: Added the constructor for backwards compatibility, but marked it as deprecated

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: See comment section. Removing this will cause backwards compatibility issues for code that calls VIPCfg.getInstance.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: See comment section. Removing this will cause backwards compatibility issues for code that calls VIPCfg.getInstance. 

In new approach, we do not allow instantiation of VIPCfg instance without product name ( like in VIPCfg.getInstance()). Instead, you should use VIpCfgFactory.getCfg(productName, true), or VIPCfgFactory.initialize(cfgBundeName, true). VIPCfgFactory.getGlobalCfg() will return null if product name is not yet set.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
Xiaochao8: By call cfg.setProductName(productName) with different product names, users can save the same config instance multiple time with different keys. Is this a problem?

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
Xiaochao8: This is confused because you created an deprecated method.

jessiejuachon: Who will be confused? Product developers will see clearly in the comment section that this is deprecated so they will not use it. They will also see clearly in the comment section which method to use instead.

Do you have any other proposal on how to keep backwards compatibility for VIPCfg.getInstance?

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: If you check cfg.setProductName, it has been deprecated and it only sets the deprecated VIPCfg.productName for backwards compatibility. It does not add the instance to configs map.

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: I know for a fact that there are products using the default constructor (e.g. Atlas).  They didn't want to use vipconfig.properties so they instantiated VIPCfg and set the properties using the setter methods with values from their application.yaml file.

src/main/java/com/vmware/vipclient/i18n/VIPCfgFactory.java
jessiejuachon: Removed!

src/main/java/com/vmware/vipclient/i18n/VIPCfg.java
jessiejuachon: Makes sense. I removed it

Xiaochao8:
Code clean up isn't clear, please use a specific PR for this change.

PR #859, opened 2020-11-03
Code clean up: unifying cache expire logic in CacheItem; deprecating old expire methods

jessiejuachon:


src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
```diff
@@ -4,6 +4,51 @@
...
+
+    public long getTimestamp() {
+        return timestamp;
+    }
+
+    protected void setTimestamp(long timestamp) {
+        this.timestamp = timestamp;
+    }
+
+    public Long getMaxAgeMillis() {
```
Xiaochao8, line 30: make this synchronized ?

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
```diff
@@ -4,6 +4,51 @@
...
+
+    protected void setEtag(String etag) {
+        this.etag = etag;
+    }
+
+    public long getTimestamp() {
+        return timestamp;
+    }
+
+    protected void setTimestamp(long timestamp) {
```
Xiaochao8, line 26: remove this or make this synchronized ?

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
```diff
@@ -4,6 +4,51 @@
...
+        return etag;
+    }
+
+    protected void setEtag(String etag) {
+        this.etag = etag;
+    }
+
+    public long getTimestamp() {
+        return timestamp;
+    }
```
Xiaochao8, line 24: make this synchronized ?

huihuiw01 (approved):

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
jessiejuachon: It has been like this before, the only change in this PR is to put common code inside the common CacheItem instead of implementing twice in MessageCacheItem and FormatCacheItem. Let's make any code logic change separately.

Also, why does it have to be synchronized? The isExpired method already has a synchronized block. Other usage of getTimestamp aside from inside isExpired do not need to be synchronized.

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
jessiejuachon: Same comment: it has been like this before, the only change in this PR is to put common code inside the common CacheItem instead of implementing twice in MessageCacheItem and FormatCacheItem. Let's make any code logic change separately.

Also, only the setters inside child classes (MessageCacheItem, PatternCacheItem, LocaleCAcheItem) use this protected setTimestamp.The setters are synchronized and have the following signature:

`public synchronized void setCacheItem(String etag, long timestamp, Long maxAgeMillis) `

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
jessiejuachon: Same comment: It has been like this before, the only change in this PR is to put common code inside the common CacheItem instead of implementing twice in MessageCacheItem and FormatCacheItem. Let's make any code logic change separately.

Also, why does it have to be synchronized? The isExpired method already has a synchronized block. Other usage of geMaxAgeMillis aside from inside isExpired do not need to be synchronized.

jessiejuachon: Hmm, let's discuss that separately. This PR is only to merge duplicate code into one.

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
jessiejuachon: Let's discuss that separately. This PR is only to merge duplicate code into one.

src/main/java/com/vmware/vipclient/i18n/base/cache/CacheItem.java
jessiejuachon: Why unnecessary? The 3 properties: etag, timestamp and maxAgeMillis have to be set in a single thread at the same time to avoid any thread collision issue.

As I said, there isn't any code logic change here in this PR. Thread-safety implementation has has been like this before this PR, so could you log a separate issue regarding any thread-safety problems you are seeing?

Xiaochao8 (approved):

PR #861, opened 2020-11-04
Issue 860: add support for .properties source files per component

jessiejuachon:


src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -62,11 +62,14 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
 				path = Paths.get(uri);
 				messages = JSONBundleUtil.getMessages(path);
 			}
-			cacheItem.setCacheItem(messages, null, System.currentTimeMillis(), null);
```
Xiaochao8, line 65: Why move this out of the try code block?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -62,11 +62,14 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
...
 		} catch (Exception e) {
 			logger.debug(e.getMessage());
 			// Do not update cacheItem
 		}
+		if (messages!=null && !messages.isEmpty())
+			cacheItem.setCacheItem(messages, null, System.currentTimeMillis(), null);
+		else
+			new LocalSrcMessagesPropsOpt(this.dto).getComponentMessages(cacheItem);
```
Xiaochao8, line 72: What's the logic behind the code? It looks not good to have this kind of logic.

Xiaochao8 (commented):
We need to confirm the requirement first.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
jessiejuachon: Only update the timestmap if the fetch was successful. However, a try catch inside FileUtil.readJson may hide a failure scenario. Checking if messages is not empty would confirm that the fetch was successful.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
jessiejuachon: Do you mean what is the issue? Please see the linked issue for details. This is for product teams that insist on using multiple .properties file instead of the VIP messages_source.json format.

jessiejuachon: By the way, this PR is not ready for review. I will re-request review once ready.

jessiejuachon: @Xiaochao8 , you do not need to approve this PR, if the solution does not look good. Do you have any suggestions? I am thinking of adding a new getMessage API with the source bundle as parameter just so we support the use case in the issue. Let me know what you think.

Xiaochao8: I'm not very clear about the requirements and your design. But it looks not good to have this kind of logic in code.

jessiejuachon: Sure, let's talk offline. Put this PR on hold.

PR #862, opened 2020-11-06
fix issue 841

huihuiw01:
1. Won't update the expire timestamp when fetch from data source failed
   2. Add data to cache when fetch successfully


src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalLocaleOpt.java
```diff
@@ -55,9 +55,6 @@ public void getSupportedLanguages(String locale, LocaleCacheItem cacheItem) {
 			logger.debug("Found the supported languages from local bundle for product [{}], version [{}], locale [{}].\n", dto.getProductID(), dto.getVersion(), locale);
 			supportedLanguageNames = JSONUtils.map2SortMap(supportedLanguageNames);
 			cacheItem.set(supportedLanguageNames, System.currentTimeMillis());
-		}else{
```
Xiaochao8, line 58: Don't need to log  a message?

Xiaochao8 (approved):

jessiejuachon (approved):

Issue #755, opened 2020-09-01
[BUG] g11n-java-client - VIP configuration for main application always applies to the shared library, except for product name and version


jessiejuachon:
**Describe the bug**
Any shared library that uses the Java client library is forced to use the VIP configurations of the root application (except for product name and version config).

**To Reproduce**
1. Use [shared library in a root application](https://github.com/vmware/singleton/tree/g11n-java-client/sample-client-app/sample-app-with-shared-lib) 
2. Add configurations for shared library that are different from the root app's config (eg. offlineResourcesBaseUrl, vipServer)
3. Notice that only the shared library's productName and version configs are used. The root app's configs are used for the rest.

**Expected behavior**
In the root application, you should be able to set the shared library's own VIP configurations properly and use it:

- productName
- version
- offlineResourcesBaseUrl
- vipServer
- initializeCache = false
- pseudo = false
- collectSource = false
- cleanCache = false
- cacheExpiredTime = 8600000
- machineTranslation = false
- defaultLocale = en

Xiaochao8:
Can you associate the PR to fix this issue?  and let the PR associate this  issues.
It's unclear which PR will fix this issue in Zenhub.

PR #869, opened 2020-11-11
Bug 868 [BUG] The offlineResourceBaseURL inside a shared lib is not being read if either main app is is run from IDE or executable jar has nested jars like in spring boot

jessiejuachon:


src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
```diff
@@ -33,22 +34,20 @@ public LocalProductOpt(BaseDTO dto) {
...
+            for (URI uri : uris) {
+                if (uri.getScheme().equals("jar")) {
+                    try (FileSystem fileSystem = FileSystems.newFileSystem(uri, Collections.<String, Object>emptyMap())) {
+                        path = fileSystem.getPath(Paths.get(uri).toString());
+                        getSupportedLocales(path, supportedLocales);
+                    }
+                } else {
+                    path = Paths.get(uri);
                     getSupportedLocales(path, supportedLocales);
                 }
```
huihuiw01, line 50: Here should only get the supported locales of main application, right? Because supported locales should oriented on main application.

If yes, when the bundle's root path of shared lib and that of main application's is the same and their supported locales are different, these code will merge shared lib's supported locale and main application's, which lead to when 'getAllComponentTranslation' method in ProductService is executed, there will be some wrong locale&component combinations.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
```diff
@@ -64,20 +63,18 @@ public void getSupportedLocales(MessageCacheItem cacheItem) {
...
+            for (URI uri : uris) {
+                if (uri.getScheme().equals("jar")) {
+                    try (FileSystem fileSystem = FileSystems.newFileSystem(uri, Collections.<String, Object>emptyMap())) {
+                        path = fileSystem.getPath(Paths.get(uri).toString());
+                        getComponents(path, components);
+                    }
+                } else {
+                    path = Paths.get(uri);
                     getComponents(path, components);
                 }
```
huihuiw01, line 77: Here should get the merged result of shared lib's component list and main application's so that all of them can be cached at main application start up, right?

But when the bundle's root path of shared lib and that of main application's is different, you can only get main application's component list.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -48,19 +49,19 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
 		try {
 			String filePath = FormatUtils.format(OFFLINE_RESOURCE_PATH, dto.getComponent(), bestMatch.toLanguageTag());
 			Path path = Paths.get(VIPCfg.getInstance().getOfflineResourcesBaseUrl(), filePath);
```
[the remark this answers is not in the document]
jessiejuachon: Hi Huihui, I think you are talking about a different issue/PR which will be resolved separately.
issue:https://github.com/vmware/singleton/issues/755 
PR: https://github.com/vmware/singleton/pull/858

There is no change in this PR regarding configuring the offlineResourceBaseURL. This bug/PR assumes that offlineResourceBaseURL is still configured in the main VIPCfg config, globally shared across main and "sub" Instances. In other words, the issue above is not related to this. 

The bug is that only the offlineResourceBaseURL folder (again, configured globally in main cfg) inside the product's main file system is loaded. The offlineResourceBaseURL folder inside the subinstance (shared lib) jar is not read. The expected behavior is that it will also load the bundles inside the offlineResourceBaseURL folder in the shared library jars. In other words, if the the shared library jar carries the translations inside it, the translation should be loaded by the root application.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: Yes, you are right. For now, getSupportedLocales will include all locales under all components under the offlineResourceBaseURL (whether inside main app or dependency jars). However, a wrong locale&component combination may happen even without a shared lib. For instance, 2 components in the main app may have different sets of bundles at some point (e.g. component A's translations are ready but component B's translations are not ready yet), so they have different supported locales. This also give a false positive in some combinations like what you said, and so will still try to fetch for the bundle even if the locale does not exist under the component.

We have been ignoring this issue even before this PR. We should improve this separately. In my opinion, getSupportedLocales should be fetched and cached per component, which will resolve the issue. This should apply to components in both the main app and shared libs. This has to be done outside of this PR.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: You are right again! This is exactly why we should be able to configure a different offlineResourceBaseURL (and other properties) for shared libs. This is tracked in issue: https://github.com/vmware/singleton/issues/755

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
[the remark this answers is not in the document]
jessiejuachon: On a side note that is not related to the scope of this PR, we will resolve your concern about being able to configure separate VIPCfg properties per product/shared lib in issue:https://github.com/vmware/singleton/issues/755. You are headed towards the right direction though. VIPCfg.getInstance().. is not correctly used. Once we remove this singleton pattern implementation through PR: https://github.com/vmware/singleton/pull/858, instead of VIPCfg.getInstance(), we will start using VIPCfgFactory.getCfg(productname) for product-specific cases such as L3, and VIPCfgFactory.getCfg() for non-product-specific cases such as for patterns. See the samples in the PR and my reply to your comments there for your reference. Again, this PR is not related to that.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
huihuiw01: I think 2 components in the main app have different sets of bundles can only happen at development stage, rarely happen at production stage, so wrong locale&component combination rarely happen at production stage, too. So before this PR we can ignore the issue. 

But after this PR, wrong locale&component combination may happen at production stage. So there may be necessary to improve this.

I think there maybe no necessary to getSupportedLocales per product component, add {product}/{version} floder in local bundle path is enough, how do you think?

huihuiw01 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
huihuiw01: But I think issue #755 only track the problem that main app and shared libs are able to use different config.

The problem still need to be resolved here is: get the merged result of shared lib's component list and main application's. This problem can be resolved either in PR #858 or in a new PR.

jessiejuachon: You are right. Thats still needs to be resolved after 858. PR 858 is the first step towards that.

For this PR, assume that the main and shared lib's product name and offline bundle paths are exactly the same. There is no problem in a fat executable jar, but there is a problem when using IDE (or whenever the shared lib is a nested jar).

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
huihuiw01, line 50: Both if block and else block call 'getSupportedLocales(path, supportedLocales);' , so it's better extract this line out of if-else block.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
huihuiw01, line 77: Both if block and else block call 'getComponents(path, components);' , so it's better extract this line out of if-else block.

huihuiw01 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: I had the same thought even before this PR. I tried to extract it, but I got a FileSystemNotFound exception from the first block. getSupportedLocales must be inside the try block. After the try block, the file system had been closed already.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: Same reply as above. This was not changed in this PR.

huihuiw01 (dismissed):

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: No we cannot ignore the issue. It could happen that there are multiple shared libraries (hence, many URIs) using the same product-version, but different component-locales set in each shared library. This happens in production. If we do not get all of those URIs, we do not get the complete.

Again, let's resolve the issue of getting supported locales and components by product-version in #870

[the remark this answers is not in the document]
jessiejuachon: Did the same mistake which causes a FileSystemNotFoundException. Because line 43 opens the file system, getSupportedLocales must be executed inside the try, before the file system closes.

jessiejuachon: > I agree, the correct solution should be getting supported locales by product and version, not get them from all product-versions.

Yes, I agree. We will work on that separately.

Xiaochao8 (approved):

linr211 (commented):
I have no question to this PR's code change, but I am wondering if this solution makes sense to Singleton java client or not:
 The Singleton jar replies on the offlineResourceBaseURL path which belongs to the product, meaning the offlineResourceBaseURL in potential is not reachable, extremely if there's a wrapper to the  offlineResourceBaseURL's bundle then Singleton jar can't read it (this project's bundle could locate at a fat jar inside, other projects may locate at a zip file inside, even at DB), so to my understanding this is a workaround solution, right?
 Another potential problem is: if there's new translation need updates, it requires to update the product's build, that make Singleton solution is not independent and we are back to traditional way.

linr211 (approved):
fine with the code change, but still have doubts on the solution.

jessiejuachon:
> I have no question to this PR's code change, but I am wondering if this solution makes sense to Singleton java client or not:
> The Singleton jar replies on the offlineResourceBaseURL path which belongs to the product, meaning the offlineResourceBaseURL in potential is not reachable, extremely if there's a wrapper to the offlineResourceBaseURL's bundle then Singleton jar can't read it (this project's bundle could locate at a fat jar inside, other projects may locate at a zip file inside, even at DB), so to my understanding this is a workaround solution, right?
> Another potential problem is: if there's new translation need updates, it requires to update the product's build, that make Singleton solution is not independent and we are back to traditional way.

Very good points, Colin. See how these concerns are addressed in PR https://github.com/vmware/singleton/pull/878

lyiyu66:
The issue is same as issue #https://github.com/vmware/singleton/issues/882 and it has been fixed with commit 3678872.

PR #875, opened 2020-11-17
change java client dependent pattern lib version to fix App-check failure

huihuiw01:
JIRA Sub-task: https://jira.eng.vmware.com/browse/VIP-2691

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,7 +46,7 @@ dependencies {
 	compile("com.googlecode.json-simple:json-simple:1.1.1") {
 	   exclude group: 'junit'
 	}
-	compile group: 'com.vmware.singleton', name: 'singleton-i18n-patterns-core', version: '+'	
+	compile group: 'com.vmware.singleton', name: 'singleton-i18n-patterns-core', version: '0.+'
 	compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.26'
 
 	testCompile group: 'org.slf4j', name: 'slf4j-log4j12', version: '1.7.26'
--- a/build.gradle
+++ b/build.gradle
@@ -46,7 +46,7 @@ dependencies {
 	compile("com.googlecode.json-simple:json-simple:1.1.1") {
 	   exclude group: 'junit'
 	}
-	compile group: 'com.vmware.singleton', name: 'singleton-i18n-patterns-core', version: '0.+'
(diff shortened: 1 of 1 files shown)


nannany00 (dismissed):

build.gradle
```diff
@@ -46,7 +46,7 @@ dependencies {
 	compile("com.googlecode.json-simple:json-simple:1.1.1") {
 	   exclude group: 'junit'
 	}
-	compile group: 'com.vmware.singleton', name: 'singleton-i18n-patterns-core', version: '+'	
+	compile group: 'com.vmware.singleton', name: 'singleton-i18n-patterns-core', version: '0.+'
```
Xiaochao8, line 49: Does 0.+ mean any version of 0.x?
If so, how about 0+? Does this mean any version bigger than 0?

huihuiw01: Yes.

According to nannan, change like this still can't resolve the problem, looks like we can't have '+' anywhere in pom file.
So I change the version to a fixed one.

build.gradle
```diff
@@ -46,7 +46,7 @@ dependencies {
 	compile("com.googlecode.json-simple:json-simple:1.1.1") {
 	   exclude group: 'junit'
 	}
-	compile group: 'com.vmware.singleton', name: 'singleton-i18n-patterns-core', version: '+'	
+	compile group: 'com.vmware.singleton', name: 'singleton-i18n-patterns-core', version: '0.5.1'
```
Xiaochao8, line 49: Please check this <a href='https://docs.gradle.org/current/userguide/single_versions.html'>Declaring Versions and Ranges</a>
A Maven-style version range: e.g. [1.0,), [1.1, 2.0), (1.2, 1.5]

The [ and ] symbols indicate an inclusive bound; ( and ) indicate an exclusive bound.

When the upper or lower bound is missing, the range has no upper or lower bound.

The symbol ] can be used instead of ( for an exclusive lower bound, and [ instead of ) for exclusive upper bound. e.g ]1.0, 2.0[
```suggestion
	compile group: 'com.vmware.singleton', name: 'singleton-i18n-patterns-core', version: '[0,)'
```

Xiaochao8 (changes requested):
Suggest change to [0,)

tigershi (approved):

linr211 (approved):

Xiaochao8 (approved):

Issue #877, opened 2020-11-17
[ENHANCEMENT and BUG] Non-blocking getSupportedLocales when fetching messages


jessiejuachon:
**Is your feature request related to a problem? Please describe.**
There are 2 issues:
1. Message fetch fails if getSupportedLocales failed, even if messages are available in remote/local dataSource. 
This should not be the case because getting the list of supported locales is only primarily used for caching and performance.
2. Message fetch is delayed if supported locales hasn't been cached.

**Describe the solution you'd like**
Even if getSupportedLocales had failed (e.g. VIP service was temporarily down), message fetch should not be blocked.
Even if the set of supported locales is not yet in cache (e.g. supported locales cache is not initialized (i.e initializeCache=false)), message fetch should not have to wait. getSupportedLocales should be done in a separate concurrent thread.

lyiyu66:
The issue and specific Windows issue have been fixed with commit 3678872.

PR #878, opened 2020-11-17
Non-blocking getSupportedLocales + Use getResourceAsStream to read resource bundle

jessiejuachon:
For the reviewers: 

Issues described in https://github.com/vmware/singleton/issues/882:

1. The java client library tries to walk the file system when getting the list of supported locales. This fails when the file system is in a nested jar. Such is the case in Spring boot applications where the offline bundles directory may exist inside dependency jars which under BOOT-INV/libs. This is related to #877

2. The java client library tries to access the file system when reading the bundle (code is in LocalMessagesOpt.getComponentMessages). This fails when the file system is in a nested jar.

**All files changed in this PR are for # 1 above, except for change sin LocalMessagesOpt.java which is for # 2.**


src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -111,20 +130,42 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+			}
+		};
+		new Thread(runnable).start();
+	}
+
 	private void doLocaleMatching() {
-		Locale matchedLocale = LocaleUtility.pickupLocaleFromList(new ProductService(dto).getSupportedLocales(), Locale.forLanguageTag(dto.getLocale()));
-		if (matchedLocale != null) // Requested locale matches a supported locale (eg. requested locale "fr_CA matches supported locale "fr")
+		//Refresh the cache of supported locales as needed in a separate thread (non-blocking).
+		refreshSupportedLocalesTask();
```
huihuiw01, line 164: I think before starting this task you need check if the supported locales list in cache by calling 'new ProductService(dto).getSupportedLocales(false)', if not in cache you need start this task, otherwise you needn't refresh cache.

Without the check, each call of 'getMessage' method will trigger the cache refresh of supported locale list actually which is needn't.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
```diff
@@ -35,27 +35,28 @@ public LocalProductOpt(BaseDTO dto) {
...
             Path path = Paths.get(VIPCfg.getInstance().getOfflineResourcesBaseUrl());
 
             URI uri = Thread.currentThread().getContextClassLoader().
                     getResource(path.toString()).toURI();
 
             if (uri.getScheme().equals("jar")) {
-                try (FileSystem fileSystem = FileSystems.newFileSystem(uri, Collections.<String, Object>emptyMap())) {
-                    path = fileSystem.getPath(path.toString());
-                    getSupportedLocales(path, supportedLocales);
+                synchronized (LocalFileSystem.getInstance()) {
```
huihuiw01, line 44: What's the purpose of this synchronized block?

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -49,8 +50,11 @@ private void refreshCacheItem(final MessageCacheItem cacheItem, Iterator<DataSou
 			return;
 		}
 
-		long timestampOld = cacheItem.getTimestamp();
 		DataSourceEnum dataSource = msgSourceQueueIter.next();
+		if (!proceed(dataSource))
+			refreshCacheItem(cacheItem, msgSourceQueueIter); // Try the next dataSource in the queue
```
huihuiw01: It will continue execute the following code after this line finish.
So either add return statement after this line or put it into if-else block.

huihuiw01 (changes requested):

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -49,25 +50,44 @@ private void refreshCacheItem(final MessageCacheItem cacheItem, Iterator<DataSou
...
-		}
-		dataSource.createMessageOpt(dto).getComponentMessages(cacheItem);
-		long timestamp = cacheItem.getTimestamp();
-		if (timestampOld == timestamp) {
-			logger.debug(FormatUtils.format(ConstantsMsg.GET_MESSAGES_FAILED, dto.getComponent(), dto.getLocale(), dataSource.toString()));
-		}
-		dto.setLocale(localeOrig);
+		if (!proceed(dataSource)) {
+			refreshCacheItem(cacheItem, msgSourceQueueIter); // Try the next dataSource
+		} else {
```
Xiaochao8, line 56: else is unnecessary because there is no code after else block. Removing it makes easy to read.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -49,25 +50,44 @@ private void refreshCacheItem(final MessageCacheItem cacheItem, Iterator<DataSou
...
+	 * @return 'true' for either of the following cases. Otherwise, false (locale not supported in data source).
+	 * <ul>
+	 * 	<li>the dataSource's set of supported locales is not in cache. If the list is not in cache, it should not block refreshCacheItem</li>
+	 * 	<li>the requested locale is found in the data source's cached set of supported locales.</li>
+	 * </ul>
+	 */
+	private boolean proceed(DataSourceEnum dataSource) {
+		ProductService ps = new ProductService(dto);
+		Set<Locale> supportedLocales = ps.getSupportedLocales(false, dataSource);
+		logger.debug("supported languages: [{}]", supportedLocales.toString());
```
Xiaochao8, line 87: No need to call "toString()" method as formatting and string conversion is done by the Formatter.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -72,24 +72,41 @@ public ProductService(BaseDTO dto) {
...
+     * @param withCacheRefresh If true, it recursively applies data source fallback mechanism in case of failure.
      * @return list of locales of the product specified in the dto object, or an empty list in case of failure to retrieve from any data source.
      */
-    public Set<Locale> getSupportedLocales() {
-        return langTagtoLocaleSet(getSupportedLanguageTags());
+    public Set<Locale> getSupportedLocales(boolean withCacheRefresh) {
+        return langTagtoLocaleSet(getSupportedLanguageTags(withCacheRefresh));
+    }
+
+    public Set<Locale> getSupportedLocales(boolean withCacheRefresh, DataSourceEnum dataSource) {
```
Xiaochao8, line 89: I think it's not good to bring so many parameters to method and so many similar methods. This makes it difficult to read and hard to maintain.
Please improve the design.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -72,24 +72,41 @@ public ProductService(BaseDTO dto) {
...
-     * Retrieves the list of locales of a product. It recursively applies data source fallback mechanism in case of failure.
+     * Retrieves the list of locales of a product.
      *
+     * @param withCacheRefresh If true, it recursively applies data source fallback mechanism in case of failure.
      * @return list of locales of the product specified in the dto object, or an empty list in case of failure to retrieve from any data source.
      */
-    public Set<Locale> getSupportedLocales() {
-        return langTagtoLocaleSet(getSupportedLanguageTags());
+    public Set<Locale> getSupportedLocales(boolean withCacheRefresh) {
+        return langTagtoLocaleSet(getSupportedLanguageTags(withCacheRefresh));
```
Xiaochao8, line 86: It's not good to use this method everywhere. We should use only one type - set or list - internally.

Xiaochao8 (changes requested):

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -111,20 +131,43 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+			}
+		};
+		new Thread(runnable).start();
+	}
+
 	private void doLocaleMatching() {
-		Locale matchedLocale = LocaleUtility.pickupLocaleFromList(new ProductService(dto).getSupportedLocales(), Locale.forLanguageTag(dto.getLocale()));
-		if (matchedLocale != null) // Requested locale matches a supported locale (eg. requested locale "fr_CA matches supported locale "fr")
+		//Refresh the cache of supported locales as needed in a separate thread (non-blocking).
+		refreshSupportedLocalesTask();
```
Xiaochao8, line 164: By our design, the refreshment should be done at each time getting supported locales rather than here.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -72,24 +72,41 @@ public ProductService(BaseDTO dto) {
...
+        return langTagtoLocaleSet(getSupportedLanguageTags(withCacheRefresh, dataSource));
     }
 
-    public Set<String> getSupportedLanguageTags() {
+    private Set<String> getSupportedLanguageTags(boolean withCacheRefresh) {
         Iterator<DataSourceEnum> msgSourceQueueIter = VIPCfg.getInstance().getMsgOriginsQueue().iterator();
         Set<String> supportedLangTags = new HashSet<>();
         while(msgSourceQueueIter.hasNext()) {
-            supportedLangTags.addAll(getSupportedLanguageTags(msgSourceQueueIter.next()));
+            supportedLangTags.addAll(getSupportedLanguageTags(withCacheRefresh, msgSourceQueueIter.next()));
```
Xiaochao8, line 97: We should cache this instead of merging every time.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalFileSystem.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2019-2020 VMware, Inc.
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package com.vmware.vipclient.i18n.messages.api.opt.local;
+
+public class LocalFileSystem {
```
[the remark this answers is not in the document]
jessiejuachon: This is used as a lock in LocalProductOpt. FileSystems.newFilesystem is not thread safe. This is evident when I put getSupportedLocales (refresh) in a separate thread. FileSystemAlreadyExistsException exception is thrown if another thread has opened the Opened the file system, making the current thread fail. We do not want the current thread to fail. Instead, we want it to wait for the other thread to finish so it could access the file system properly.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: If you trace this code, it will eventually call `ProductService.getSupportedLanguageTags(DataSourceEnum)`. In there, it returns the cached data if it exists in cache and is not expired.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: Pasting my reply to Xiaochao above:

_"FileSystems.newFilesystem is not thread safe so we want only 1 thread to be able to do it at a time. FileSystemAlreadyExistsException exception is thrown if another thread is simultaneously accessing the file system, making the current thread fail. We do not want the current thread to fail. Instead, we want it to wait for the other thread to finish so it could access the file system properly. To test this, have multiple threads try to access the file system (e.g. 100 calls that will get the list of supported locales at the same time.)"_

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: I had updated this yesterday. Perhaps you did not see?

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: What is your suggested design then? Let's be more clear so we could move faster.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: They are both Set type. List was not used. This method converts from Set<String> language tags to Set<Locale> locales

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -72,24 +72,41 @@ public ProductService(BaseDTO dto) {
...
         }
         return supportedLangTags;
     }
 
+    private Set<String> getSupportedLanguageTags(boolean withCacheRefresh, DataSourceEnum dataSource) {
+        if (withCacheRefresh)
+            return getSupportedLanguageTags(dataSource);
```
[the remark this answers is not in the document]
jessiejuachon: the behavior is the same as before. If cacheItem is expired, `getSupportedLanguageTags(dataSource)` returns the expired cacheItem and refreshes the cacheItem in a separate thread. If the cacheItem is not in cache yet and populating the cache fails, it returns an empty Set. No changes here.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: That is not true. See Locale Matching part of the design spec: https://confluence.eng.vmware.com/pages/viewpage.action?pageId=530637182#SingletonClientLibrary-Caching,FallbackMechanism,andOtherFeatures-L10nLocaleMatching
If "List of supported locales in cache" = N, it should return the requested originally requested locale.

In the Query part of the design spec: https://confluence.eng.vmware.com/pages/viewpage.action?pageId=530637182#SingletonClientLibrary-Caching,FallbackMechanism,andOtherFeatures-L10nQuery, you will see that there is no judgment box for "L10n Locale Matching". The path just continues no matter what the result of the locale matching is (non-blocking)

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: Caching a combined set will duplicate the data in memory, so the footprint will unnecessarily increase. Note that the separated sets (per data source) are also required.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: See the updated code. I have cleaned up some methods.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: See the updated code. I have moved this utility method to LocaleUtilityand used it for the call to LocaleUtility.pickupLocaleFromList only.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalFileSystem.java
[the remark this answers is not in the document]
jessiejuachon: I have changed this and moved the handling of thread safety in ProductService. Please check synchronized blocks in ProductService.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: I have changed this and moved the handling of thread safety in ProductService. Please check synchronized blocks in ProductService.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalFileSystem.java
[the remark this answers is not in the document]
Xiaochao8: Because we just read the file so lock is unnecessary.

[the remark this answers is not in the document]
Xiaochao8: The file system is corresponding to uri, so it shouldn't to have only one lock for all file systems.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -109,22 +132,28 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
 				cacheItem = getFallbackLocaleMessages(fallbackLocalesIter);
 		}
+
 		return cacheItem;
 	}
 
 	private void doLocaleMatching() {
-		Locale matchedLocale = LocaleUtility.pickupLocaleFromList(new ProductService(dto).getSupportedLocales(), Locale.forLanguageTag(dto.getLocale()));
-		if (matchedLocale != null) // Requested locale matches a supported locale (eg. requested locale "fr_CA matches supported locale "fr")
+		dto.setLocale(LocaleUtility.fmtToMappedLocale(dto.getLocale()).toLanguageTag());
```
Xiaochao8, line 144: If this is necessary. Suggest to do this when create dto object.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -72,44 +73,75 @@ public ProductService(BaseDTO dto) {
         return components;
     }
 
+    private Set<String> getSupportedLocales() {
+        return getSupportedLocales(true);
+    }
+
+    public Set<String> getCachedSupportedLocales() {
```
huihuiw01, line 80: There are 3 'getSupportedLocales' methods and 2 'getCachedSupportedLocales' methods, and they call each other, and seems 2 'getCachedSupportedLocales' are not the relation of overload, it's a little difficult to understand the relationship of these methods, it's better optimize the methods' name.

huihuiw01 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
huihuiw01: - Caching a combined set should reduce the data in memory, I can't understand how duplicate?
- Since the locale list and component list data size is small, so the footprint won't increase so much.
- Also don't understand why separated sets (per data source) are also required?

Just have these questions, and they may not have relation to 882, so shouldn't block this PR merged.
If need optimize, of course this can be done in a new PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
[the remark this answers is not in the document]
huihuiw01: I also agree cache locale list when initialization no matter initializeCache is true or false, since it may reduce the chance of locale list fetching be triggered by message fetching which can improve performance at message fetching.
Further more, we may remove fetch locale list task code at message fetching.

This can be done in a new PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: - Duplicate because cache will have: 

1. cache of offline + online locales 
2. cache of offline locales 
3. cache of online locales
# 1 above is a duplicate of # 2 + # 3

- Not so much of an increase, but why the need to increase it? Combining 2 and 3 at runtime is O(1) time complexity anyway. I think either way, is fine.

- See how ComponentService.proceed uses ProductService.getCachedSupportedLocales(dataSource). It checks if the locale currently being requested is supported in the **specific dataSource**.  This is because the list of supported locales may be different between offline and online at runtime. We had discussed some use cases some time ago: if new locales are added to remote VIP service at runtime, or if new message bundles are added to SAAS product before updating VIP service.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -109,22 +132,28 @@ public MessageCacheItem getMessages(Iterator<Locale> fallbackLocalesIter) {
...
+
+		//Refresh the cache of supported locales as needed in a separate thread (non-blocking).
+		ps.refreshSupportedLocalesTask();
+
+		//Match against list of supported locales that is already in the cache
+		Set<Locale> supportedLocales = LocaleUtility.langTagtoLocaleSet(ps.getCachedSupportedLocales());
+		Locale matchedLocale = LocaleUtility.pickupLocaleFromList(supportedLocales, Locale.forLanguageTag(dto.getLocale()));
+		if (matchedLocale != null) { // Requested locale matches a supported locale (eg. requested locale "fr_CA matches supported locale "fr")
 			dto.setLocale(matchedLocale.toLanguageTag());
+		}
```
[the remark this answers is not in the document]
jessiejuachon: Looks pretty much the same to me. Based on the flow you described
1. ComponentService will get supported locales from product service
2. The product service returns an empty set if cache of locales is not found.
3. The Product Service will refresh the cache if needed (Note: must be in a separate thread).
4. If supported locales set isn't null/empty, then pick up the locale from them.

# 1 This PR already does it when ComponentService.doLocaleMatching calls ps.getCachedSupportedLocales(). 
# 2 In this PR,  ProductService.getCachedSupportedLocales() in # 1 returns an empty set if not found in cache. 
# 3 Sure, I can call ProductService.refreshSupportedLocalesTask inside ProductService.getCachedSupportedLocales instead of here.
# 4 This Pr does it when ComponentService.doLocaleMatching sets dto.locale with the matchedlocale: `dto.setLocale(matchedLocale.toLanguageTag());`

The only added logic to the flow is ComponentService.proceed, which is only called when message bundle is not yet in cache or is expired. This is for performance enhancement.  ComponentService.proceed calls ps.getCachedSupportedLocales(dataSource) to check if the requested locale (dto.getLocale()) is supported in that specific data source. This is so that if it is not supported in that data source, it won't even try to fetch the message bundle. This is important especially if that invalid fetch is to a remote VIP service which will affect performance.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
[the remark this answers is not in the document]
jessiejuachon: Please review the other issue linked to this: https://github.com/vmware/singleton/issues/877: 

We do not need getCachedXXX for other data.

getCachedSupportedLcoales is an exception because we use it to pre-determine if getComponentMessages must even proceed or not. This was done only for performance purposes, so that if locale is not supported, it won't even try to fetch the bundle and directly serve fallback locale bundle. However, if getSupportedLocales was not previously cached or if it previously failed, it should not block getComponentMessages. The client should just directly send  the call to getComponentMessages.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -72,44 +73,75 @@ public ProductService(BaseDTO dto) {
...
+     * @return The set of locales supported in the given data source.
+     */
+    public Set<String> getSupportedLocales(DataSourceEnum dataSource) {
         CacheService cs = new CacheService(new MessagesDTO(dto));
         MessageCacheItem cacheItem = cs.getCacheOfLocales(dataSource);
         if (cacheItem != null) {
-            if (cacheItem.isExpired())
-                refreshLocalesCacheItemTask(cacheItem, dataSource);
+            if (cacheItem.isExpired()) {
+                synchronized (cacheItem) { // Allow only 1 thread to refresh the cacheItem at a time.
```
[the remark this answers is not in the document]
jessiejuachon: Yes, and that is why inside the synchronized block, another `if (cacheItem.isExpired())` is done so that it will not proceed with the update if previous thread has already refreshed the cacheItem.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -72,44 +73,75 @@ public ProductService(BaseDTO dto) {
...
+                    if (cacheItem.isExpired())
+                        refreshLocalesCacheItemTask(cacheItem, dataSource);
+                }
+            }
             return cacheItem.getCachedData().keySet();
         } else {
-            cacheItem = createLocalesCacheItem(dataSource);
+            // Allow only 1 thread to create the cacheItem.
+            synchronized (TranslationCacheManager.getInstance().getCache(VIPCfg.CACHE_L3)) {
```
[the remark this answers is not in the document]
jessiejuachon: I agree. Any suggestions on what object to use as a lock that will be specific to the item that does not exist yet?

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -72,44 +73,75 @@ public ProductService(BaseDTO dto) {
...
+     * @param refreshCache If true, it will trigger a cache populate or refresh as necessary before returning.
+     *                         If false, it will return the data from the cache as is, or an empty Set if not in cache.
+     * @return The set of supported locales.
      */
-    public Set<Locale> getSupportedLocales() {
-        return langTagtoLocaleSet(getSupportedLanguageTags());
-    }
-
-    public Set<String> getSupportedLanguageTags() {
+    private Set<String> getSupportedLocales(boolean refreshCache) {
```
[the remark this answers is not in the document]
jessiejuachon: I disagree. Overloading methods is commonly done, especially to avoid duplicating code when there is very small difference between the 2 methods. I have seen this bad practice of duplicating methods and code in other places in our java client. Let's try to avoid this. This means having to maintain 2 identical code, having to make sure that if you update one, the other/s have to be updated too with the exact same change, etc.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
```diff
@@ -49,25 +50,47 @@ private void refreshCacheItem(final MessageCacheItem cacheItem, Iterator<DataSou
...
+	private boolean proceed(DataSourceEnum dataSource) {
+		ProductService ps = new ProductService(dto);
+		Set<String> supportedLocales = ps.getCachedSupportedLocales(dataSource);
+		logger.debug("supported languages: [{}]", supportedLocales);
+
+		/*
+		 * Do not block refreshCacheItem if set of supported locales is not in cache (i.e. supportedLocales.isEmpty()).
+		 * This happens either when cache is not initialized, OR previous attempts to fetch the set had failed.
+		 */
+		return (supportedLocales.isEmpty() || supportedLocales.contains(dto.getLocale()) || VIPCfg.getInstance().isPseudo());
```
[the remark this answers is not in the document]
jessiejuachon: This has been there since before this PR. I just moved the if-statement judgements from ComponentService.getMessage to here, so no change here.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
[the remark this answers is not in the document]
jessiejuachon: 1. We should not force to cache this at app start up. This is why we have initializeCache property in the config. Even component messages are sometimes not initialized.
2. In some cases, this will always fail (FileSystems.newFileSystem walk fails in nested jar applications like in Atlas for Java 11 or less), so always calling getSupportedLocales that fails might affect performance a it tries to read the disk.

src/main/java/com/vmware/vipclient/i18n/messages/service/ComponentService.java
jessiejuachon: This has been there even before this PR. Let's not change it here.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
```diff
@@ -143,13 +175,16 @@ private MessageCacheItem createLocalesCacheItem(DataSourceEnum dataSource) {
         }
         return null;
     }
-    private Set<Locale> langTagtoLocaleSet (Set<String> languageTags) {
-        Set<Locale> locales = new HashSet<>();
-        if (languageTags != null) {
-            for (String languageTag : languageTags) {
-                locales.add(Locale.forLanguageTag(languageTag));
+
+    public void refreshSupportedLocalesTask() {
```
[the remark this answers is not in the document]
jessiejuachon: Moved it to ProductService,getCachedSupportedLocales too.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
jessiejuachon: I have renamed 1 method only. The overloading should be clear. Look how the methods names are self-explanatory.. descriptions seem to be repetitive of what is already obvious.

1. getSupportedLocales() - gets all supported locales (from all data sources)
2. getSupportedLocales(dataSource) - gets supported locales from the specified data source
3. getCachedSupportedLocales() - gets all supported locales (from all data sources) from cache
4. getCachedSupportedLocales(dataSource) - gets supported locales from the specified data source from cache
5. combineSupportedLocales (refreshCache) - combines the list of supported locales from all dataSources. Used in # 1 and # 3.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalFileSystem.java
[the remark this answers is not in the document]
jessiejuachon: > Because we just read the file so lock is unnecessary.

Try it with multiple threads and see the exception I said.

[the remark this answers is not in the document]
jessiejuachon: > The file system is corresponding to uri, so it shouldn't to have only one lock for all file systems.

I have moved the lock to inside LocalProductOpt.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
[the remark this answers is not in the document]
jessiejuachon: Anyway, I have removed this part out of this PR to narrow down the scope. I will include it in another PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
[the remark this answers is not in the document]
jessiejuachon: I have removed this part out of this PR to narrow down the scope. I will include it in another PR.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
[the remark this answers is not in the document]
jessiejuachon: I have renamed this to combineSupportedLocales. The overloading should be clear. Below are the method signatures 

1. getSupportedLocales() - gets all supported locales (from all data sources)
2. getSupportedLocales(dataSource) - gets supported locales from the specified data source
3. getCachedSupportedLocales() - gets all supported locales (from all data sources) from cache
4. getCachedSupportedLocales(dataSource) - gets supported locales from the specified data source from cache
5. private combineSupportedLocales (refreshCache) - combines the list of supported locales from all dataSources. Used in # 1 and # 3.

I have also updated javadoc for all methods.

src/main/java/com/vmware/vipclient/i18n/messages/service/ProductService.java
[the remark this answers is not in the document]
jessiejuachon: Please check ProductService. The code is now more unified.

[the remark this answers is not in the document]
Xiaochao8: It seems you created so much code just to work around an extreme situation, getSupportedLocales keeps fail. In this case,  we should remove test supported locales because stability takes precendence over performance. Let users take charge of it. Get the supported locale before getting translation or bear the delay of unsupported locale. In fact this logic has been there before for a long time.

Xiaochao8: according OO thinking, the dataSource should be responsible of its methods instead of product service.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -43,40 +43,37 @@ public JSONObject getComponentMessages() {
...
-    	}
-    	return supportedLocales;
-    }
+
+	private InputStream getInputStream() {
+		String locale = LocaleUtility.fmtToMappedLocale(dto.getLocale()).toLanguageTag();
+		while (true) {
+			String filePath = FormatUtils.format(OFFLINE_RESOURCE_PATH, dto.getComponent(), locale);
+			Path path = Paths.get(VIPCfg.getInstance().getOfflineResourcesBaseUrl(), filePath);
+			InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(path.toString());
```
Xiaochao8, line 62: Why change FileSystem to getResourceAsStream?

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalProductOpt.java
jessiejuachon: Let's narrow down the scope to deliver this faster.  Since this fix is not related to the Atlas bug, I have removed it from this PR and created this issue:  https://github.com/vmware/singleton/issues/910.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalFileSystem.java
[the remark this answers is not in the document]
jessiejuachon: Let's narrow down the scope to deliver this faster. Since this fix is not related to the Atlas bug, I have removed it from this PR and created this issue: #910.

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
jessiejuachon: Because FileSystems.newFileSyustem fails in nested jars, but getResourceAsStream does not. So the problem is just in getting the list of components or supported locales (and only in Java 11 or below), not in actually reading the resource bundle.

jessiejuachon:
@Xiaochao8  

> "As huihui said, the additional footprint isn't much because they are just a very few strings. But will improve performance greatly. Don't need to combine two set anymore."

As I said, pre-loading anything is not good practice. Actually, pre-loading the list of supported locales will affect performance of application start up. How does it improve performance? Again, what's wrong with getting this only when needed in a separate thread?

jessiejuachon:
@Xiaochao8 

> It seems you created so much code just to work around an extreme situation, getSupportedLocales keeps fail. In this case, we should remove test supported locales because stability takes precendence over performance. Let users take charge of it. Get the supported locale before getting translation or bear the delay of unsupported locale. In fact this logic has been there before for a long time.

This "so much code" is NOT just to "workaround an extreme situation". This is the proper behavior: The first step (getSupportedLocales) that is only used for performance purposes should not block the second step (getComponentMessages). This is regardless of whether FileSystems walk fails or not. ANY getSupportedLocales failure (offline or online mode) should not block getComponentMessages. For instance, getSupportedLocales may from remote may fail too (e.g. temporary network issue), but getComponentMessages may still succeed. 

This is what this PR is about. It proves that even if getSupportedLocales step fails, messages can still be fetched. 

The issue of FileSystem walk not working happens whenever the application has shared libs and are using Spring boot (or any platform that packages with nested jars instead of 1 fat jar). So this will be quite common. You are right, we should fix this issue for stability, but our hands are tied for now because FileSystem.newFilesSystem in nested jars is not supported in Java 11. We can explore other options later. 

What do you mean by let users take charge of it? What can they do?

> Get the supported locale before getting translation or bear the delay of unsupported locale. In fact this logic has been there before for a long time.
You are mixing 2  issues: 
1. There is a delay if supported locales is not yet fetched 
The only added code for this problem, is createLocalesCacheItemTask -- createLocalesCacheItem in s separate thread.
2. getComponentMessages fails if first step (getSupportedLoclaes) failed.
This buggy logic has been there for a long time, but hidden for a while because we are not testing getSupportedLocales failed scenario. It was just made obvious by the Atlas bug. This is primarily what we are fixing in this PR.

jessiejuachon:
@Xiaochao8 

> according OO thinking, the dataSource should be responsible of its methods instead of product service.

getCachedSupportedLocales is certainly not a function of the dataSource. It is a function of CacheService. CacheService is utilized when it is called down the line in ProductService.processSupportedLocales:

`CacheService cs = new CacheService(new MessagesDTO(dto));
   MessageCacheItem cacheItem = cs.getCacheOfLocales(dataSource);`

Therefore, I do not think we are violating any OO concept.

huihuiw01 (approved):

Xiaochao8 (approved):
Even though there are still some places to be improved, there aren't blocker issues. Merge it now.

jessiejuachon:
> Even though there are still some places to be improved, there aren't blocker issues. Merge it now.

Thanks for merging it. I have created the following to track the FileSystems walk issue when getting supported locales in nested-jar applications: https://github.com/vmware/singleton/issues/912

PR #902, opened 2020-12-04
Support reading allow list from S3

Xiaochao8:


g11n-ws/vip-manager-l10n/src/main/java/com/vmware/l10n/utils/WhiteListUtils.java
```diff
@@ -74,4 +54,47 @@ private WhiteListUtils() {
 		return null;
 	}
 
+	public static class LocalWhitelistUtils extends WhiteListUtils {
+		@Override
+		protected String readWhitelistFile() {
```
linr211, line 59: Is there a function can read the file content from common project? Please take a check.

Xiaochao8: This is legacy code. It's better not to change it.
Please see the latest commit for the change.

linr211 (approved):

tigershi (approved):

PR #911, opened 2020-12-07
fix issue 658 about getting component list

jin-jasper:


singletonclient/SingletonClient/Implementation/SingletonUpdate.cs
```diff
@@ -222,9 +213,11 @@ private ILocaleMessages LoadOfflineLocaleBundle(string locale, string nearLocale
                 ).Split(',');
 
             List<string> componentList = _config.GetConfig().GetComponentList();
-            if (componentList == null)
+            if (componentList.Count == 0)
```
Xiaochao8, line 216: What if componentList == null?

jin-jasper: GetComponentList() doesn't return null. it must return an object.

jin-jasper: sonarQube forced me to change from returning null to returning a list with no items.

Xiaochao8 (approved):

linr211 (approved):

PR #913, opened 2020-12-08
Fix Windows bug on getComponentMessages

jessiejuachon:
Paths.get must not be use here because the defined path separator in JAR files is '/' (forward slash). Paths.get will use '\' (backslash) in Windows which results to not finding the messages_xx.json path inside the jar.


Xiaochao8 (approved):

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
```diff
@@ -67,9 +67,11 @@ public void getComponentMessages(MessageCacheItem cacheItem) {
 	private InputStream getInputStream() {
 		String locale = LocaleUtility.fmtToMappedLocale(dto.getLocale()).toLanguageTag();
 		while (true) {
-			String filePath = FormatUtils.format(OFFLINE_RESOURCE_PATH, dto.getComponent(), locale);
-			Path path = Paths.get(VIPCfg.getInstance().getOfflineResourcesBaseUrl(), filePath);
-			InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(path.toString());
+			String offlineResourcePath = VIPCfg.getInstance().getOfflineResourcesBaseUrl();
+			if(!offlineResourcePath.endsWith("/"))
```
huihuiw01, line 71: Can "/" be suitable for all systems? Should it be FileSeparator?

huihuiw01 (commented):
Looks like the change in this PR is not correct, include many changes in last PR (878).

src/main/java/com/vmware/vipclient/i18n/messages/api/opt/local/LocalMessagesOpt.java
jessiejuachon: No, that is the bug. If we use FileSeparator or Paths.get, '\' (backslash) will be used in Windows platform. However, the forward slash in jar is not platform dependent. Even if the application is run on a Windows platform, the jar files will still have forward slash. The use of Paths.get (same as FileSeparator) was causing the issue.

huihuiw01 (approved):

PR #914, opened 2020-12-09
fix issue 657 about getting locale list and set YamlDotNet to v8.1.2

jin-jasper:


linr211 (approved):

singletonclient/SingletonClient/Implementation/SingletonUpdate.cs
```diff
@@ -218,6 +218,16 @@ private ILocaleMessages LoadOfflineLocaleBundle(string locale, string nearLocale
...
                 SingletonUtil.UpdateListFromAnotherList(releaseComponents, componentList);
+
+                List<string> releaseLocales = _release.GetRelease().GetMessages().GetLocaleList();
+                if (releaseLocales.Count == 0)
+                {
+                    for (int i = 0; i < releaseComponents.Count; i++)
+                    {
+                        List<string> componentLocaleList = _config.GetExternalLocaleList(releaseComponents[i]);
+                        SingletonUtil.UpdateListFromAnotherList(releaseLocales, componentLocaleList);
+                    }
```
huihuiw01, line 229: Should you get the locale list by get the combination of locale list under each component, then use it to update the locale list?

jin-jasper: It's exactly what you said.
1. get the locale list by get the combination of locale list under each component
      List<string> componentLocaleList = _config.GetExternalLocaleList(releaseComponents[i])
2.use it to update the locale list
     SingletonUtil.UpdateListFromAnotherList(releaseLocales, componentLocaleList);

huihuiw01: But line 227 and 228 are in the for loop,
 1. Isn't componentLocaleList a new object? If yes, then it isn't the combination of locale list under each component, just that for current component.
 2. Needn't update in each iteration, update after iteration done is enough.

jin-jasper: The loop is to enumerate every component.

componentLocaleList is a new object and it is the combination of locale list under a component, 

But componentLocaleList could be different for different component. All componentLocaleList objects are combined into releaseLocales as the final result.

huihuiw01 (approved):

PR #922, opened 2020-12-14
[Devops][g11n-java-client]Add building sample-app-with-shared-lib

WisdomLuo:
Add building sample-app-with-shared-lib to unit test job for branch g11n-java-client


surite (changes requested):
Please just run this sample app in smoke test to make sure the jar can be executed without error.

WisdomLuo:
Smoke test job has included commands to start up sample-app-with-shared-lib jar

PR #931, opened 2020-12-18
Add API of getting Singleton service's version information

huihuiw01:


g11n-ws/vip-manager-i18n/src/main/resources/application.properties
```diff
@@ -4,6 +4,10 @@
 ##
 
 #build version information
+build.name=vip-i18n-manager
+build.author=VMware
+build.version=
```
linr211, line 9: The value of build.version should be overwritten in the build.gradle and the value should come from the cmd parameter of build creation, right? But i don't see how this work.

huihuiw01: Have synced this with @junw-33455432 , he will overwrite the version value when publish build, as the same with what he did for the existing builddate&buildnumber&branch, so I needn't write such task.

tigershi (approved):

Xiaochao8:
Why was a new project created for this feature? How about putting this with i18n manager together?

huihuiw01:
> Why was a new project created for this feature? How about putting this with i18n manager together?

This API is simple maybe it's ok put it together with i18n manager, but it's not good practice to put logic into Controller/API layer, this layer is for handle request parameter and response. 
Another more important reason is there maybe other similar API requirement in the feature,  which can be categorized as /about, just like /translation which includes serveral L3 APIs, then they can be added directly under this module. In other word, it's added for extendability.

linr211 (approved):
The codes are clear to me, fine to ship!

PR #948, opened 2021-01-08
add the l10n agent sync source from i10n manager s3 storage feature

tigershi:


tigershi:
because of update copyright need to close

PR #971, opened 2021-01-21
Add test case for plural rule fallback

Xiaochao8:


src/test/java/com/vmware/vip/i18n/TranslationMessageTest.java
```diff
@@ -571,4 +571,18 @@ public void testNamedArgs() {
...
+		String plural_key = "sample.plural.key1";
+		Locale locale_ar = Locale.forLanguageTag("ar");
+		
+		int artInt = 3;
+		String argString = "MyDisk";
+		Object[] msgArgs = new Object[] {artInt, argString}; //In ar, 3 belongs to 'few' type.
+
+		String message_ar = translation.getMessage(locale_ar, component, plural_key, msgArgs);
+		Assert.assertEquals("There are "+artInt+" files on disk \""+argString+"\".", message_ar);
+	}
```
Xiaochao8, line 587:
```suggestion
	public void testPluralFallback() {
		String component = "JAVA";
		String plural_key = "sample.plural.key1";
		Locale locale_ar = Locale.forLanguageTag("ar");
		
		int argInt = 3;
		String argString = "MyDisk";
		Object[] msgArgs = new Object[] {argInt, argString}; //In ar, 3 belongs to 'few' type.

		String message_ar = translation.getMessage(locale_ar, component, plural_key, msgArgs);
		Assert.assertEquals("There are "+argInt+" files on disk \""+argString+"\".", message_ar);
	}
```

tigershi (approved):

src/test/resources/mockserver/__files/body-i18n-api-v2-translation-products-JavaclientTest-versions-1.0.0-locales-en-components-JAVA-txB30.json
```diff
@@ -16,7 +16,7 @@
             "LeadTest": "[{0}] Test alert",
             "global_text_username": "User name",
             "table.host": "Host",
-            "sample.plural.key1": "{0, plural, one{There is one file on disk \"{1}\".} other{There are # files on disk \"{1}\".}}",
+            "sample.plural.key1": "{0, plural, one{There is one file on disk \"{1}\".} few {\"{1}\".} other{There are # files on disk \"{1}\".}}",
```
huihuiw01, line 19: Needn't add translation for 'few' category, as there is no 'few' category for English.

Xiaochao8: This is for test. 
In ar, 3 belongs to 'few' type.
If translation falls back to English, but plural rule doesn't. The translation will use 'few' category rather than 'other' category.

huihuiw01: As previous behavior, if translation falls back to English, plural rule will fallback to English. So the translation will use 'other' category 'en' rather than 'few' category of 'ar'.

Xiaochao8: The 'few' is necessary. If no 'few', plural rule will use 'other' always. Test case will make no sense because previously code will use 'few' category.

huihuiw01: No, at first, the translation referred to plural in each locale file should be defined according to locale's plural category, for English there are only 'one' and 'other' category, so only translation for 'one' and 'other' should be defined;

Secondly, when translation fallback to English, the locale used to get plural pattern is English, no 'few' category for English so won't get 'few' category, you can debug to verify or confirm with QE.

In addition, the sense of test case is when translation fallback to English, plural pattern also fallback to English instead of not fallback, and as a result get the translation according to English plural rule instead of locale's.

Xiaochao8: Fox example, request plural translation for 'AR' locale which has a 'few' category. Because of no translation for 'AR',  translation will fall back to default. The translation of plural behaves as below

behavior before (rule doesn't fall back): 
      Plural category will be 'few'.
    But there isn't 'few' message in English message, use 'other' instead. **Can't detect error even though rule is wrong.**

behavior now (rule falls back to English): 
      Plural category will be 'other'.
    Use 'other' message.

From above, if no 'few' message, can't detect error that rule doesn't fall back.

huihuiw01 (approved):

PR #1098, opened 2021-05-11
fix bug 936: [BUG] [Service][lite]Product allow list doesn't work in Singleton I18n Service

huihuiw01:


huihuiw01:
close this PR as the issue is not a bug, no need to fix.

PR #1134, opened 2021-06-01
add parameter validation

tigershi:


Xiaochao8 (approved):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/constants/ValidationMsg.java
```diff
@@ -26,4 +26,5 @@
 	public final static String PATTERN_NOT_VALIDE = "Incorrect pattern(only allows letter)";
 	public final static String PRODUCTNAME_NOT_SUPPORTED = "The product '%s' is NOT supported yet!";
 	public final static String LOCALENAME_NOT_SUPPORTED = "The locale '%s' is NOT supported yet!";
+	public final static String APPID_NOT_VALIDE = "Incorrect appId(only allows letter and number)";
```
huihuiw01, line 29: Looks like 'Invalid' is more precise than 'Incorrect'.

tigershi: Because other pattern validate error msg is use Incorrect

huihuiw01 (approved):

PR #1141, opened 2021-06-03
fix bug 1038: [BUG] No exception pops up when requesting non-existent key with resourceBundle parameter in getMessage() API.

huihuiw01:


tigershi (dismissed):

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -102,8 +102,6 @@ private String getMessageWithArgs(String resourceBundle, final Locale locale, fi
         ComponentService.TranslationsDTO msgsItemDTO = getMessages(locale, component, true);
         String message = msgsItemDTO.getMessages().get(key);
         if (message == null || message.isEmpty()) {
-            if (resourceBundle != null)
-                return source;
```
Xiaochao8, line 106: Why not return source but throw exception?

Xiaochao8 (approved):

tigershi (approved):

PR #1286, opened 2021-08-27
ehcache3 out of memory issue

tigershi:


linr211 (commented):
The fixed solution is just removing the cache?

Xiaochao8 (approved):

tigershi:
> The fixed solution is just removing the cache?

Yes, I have tested other method, but not work

huihuiw01 (approved):

PR #1390, opened 2021-12-02
Added CI

vm-mounika:


surite (commented):
please fix CI first

.github/workflows/main.yaml
```diff
@@ -0,0 +1,99 @@
+name: main
```
nannany00, line 1: Could you change the name to "Singleton-Angular-Client", then we can distinguish it in Actions well?

vm-mounika: Updated

surite (approved):

nannany00 (approved):

PR #1432, opened 2022-01-12
 [i18n-manager] change parameter source to optional in string-based api

linr211:
Changes: 
    1. change parameter source to optional in string-based api
    2. remove the source output in the response as it will be large
    3. improve messages


tigershi (approved):

Xiaochao8 (approved):

g11n-ws/modules/md-service-i18n-l3/src/main/java/com/vmware/vip/core/messages/service/string/StringService.java
```diff
@@ -62,48 +62,21 @@ public StringBasedDTO getStringTranslation(
...
+		StringBasedDTO strDTO = null;
 
 		// get translation
 		try {
 			translation = getString(comDTO, key);
-		} catch (L3APIException e) {
-			logger.error(e.getMessage(), e);
-		}
-
-		// get en-US string
```
tigershi, line 73: there no locale fall back

PR #1438, opened 2022-01-13
upgrade the gradle scripts

tigershi:
update autotest/service/i18n-service/APITest/build.gradle
 update autotest/service/l10n-service/APITest/build.gradle


surite (commented):
please explain more about why you do this change, and introduction about 'from configurations.runtimeClasspath', is this a gradle version update? it is better to have a more clear title.

nannany00 (approved):

tigershi:
In new version gradle use configurations.runtimeClasspath not configurations.compile

PR #1467, opened 2022-01-27
[G11n ruby client]improve empty check and unit test

linr211:
improve empty check and unit test


jin-jasper (approved):

tigershi (approved):

lib/sgtn-client/api/translation.rb
```diff
@@ -15,6 +15,9 @@ def self.getString(component, key, locale)
         str = getTranslation(component, key, locale)
         if str.nil?
           str = SgtnClient::Source.getSource(component, key, SgtnClient::Config.configurations.default)
+          if str.nil?
+            SgtnClient.logger.error "Can't find the key " + key + " in source path!"
```
Xiaochao8, line 19: Suggest add single quote around key, log component if possible. So we know the necessary easily.

lib/sgtn-client/api/translation.rb
```diff
@@ -23,6 +26,10 @@ def self.getString_p(component, key, plural_args, locale)
...
```
Xiaochao8, line 30: Suggest add single quote around key, log component if possible. So we know the necessary easily.

lib/sgtn-client/api/translation.rb
```diff
@@ -48,7 +58,7 @@ def self.getStrings(component, locale)
           items = {}
           s = SgtnClient::Source.getSources(component, default)
           if s.nil?
-            SgtnClient.logger.error "Can't find the component " + component + "in source path!"
+            SgtnClient.logger.error "Can't find the component " + component + " in source path!"
```
Xiaochao8, line 61: Suggest add single quote around component

Xiaochao8 (approved):

Issue #1478, opened 2022-02-07
[BUG] [Ruby Client] Client library has hardcoded "default" locale.


jessiejuachon:
**Describe the bug**
 The client is not fetching a list of fallback locales and source localle from client configuration file. Instead, a "default" locale is hardcoded.

**To Reproduce**
1. Use the ruby client in a sample application.
2. Error: Notice that in the client configuration file, there is no config to set the following:
- list of fallback locales (ordered in priority)
- sourece locale

**Expected behavior**
The above two configurations must be in the client config file. These are to be used for the client library workflow logic below.

When target locale is not supported, invalid or nil:
a. If list of fallback locales is NOT configured in the client config file, getString must return the source string.
b. If list of fallback locales is configured in the client config file, getString must use the next fallback locale from the list as target locale, and return the translation. It must go through the list until it is able to return a translation in a fallback locale. If at the end of the list and all fallback locales failed, it must return the source string.

linr211:
How to define the list of fallback locales, could you provide an example?

PR #1495, opened 2022-02-10
Fixing issue 1476 - An error is raised when locale argument is nil

jessiejuachon:


Xiaochao8 (approved):

lib/sgtn-client/api/translation.rb
```diff
@@ -86,6 +86,7 @@ def self.getTranslation(component, key, locale)
       end
 
       def self.get_cs(component, locale=nil)
+        locale = locale.to_s
```
Xiaochao8, line 89: An exception will throw when locale is nil.

jessiejuachon: No, it won't. Test nil.to_s and it didn't raise an exception

spec/unit/onclient_spec.rb
```diff
@@ -16,6 +16,11 @@
       expect(SgtnClient::Translation.getString("JAVA", "helloworld", "en")).to eq 'Hello world'
     end
 
+    it "GET_INVALID_LOCALE" do
+      expect(SgtnClient::Translation.getString("JAVA", "helloworld", locale: nil)).to eq 'Hello world'
+      expect(SgtnClient::Translation.getString("JAVA", "helloworld", locale: 'garbage')).to eq 'Hello world'
```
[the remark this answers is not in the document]
jessiejuachon: > Since '`locale`' has a default value 'nil', `locale` parameter can be removed from input arguments.

I do not get this? Where do you want to remove 'locale' argument from?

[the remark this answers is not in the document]
jessiejuachon: Do you mean `expect(SgtnClient::Translation.getString("JAVA", "helloworld")).to eq 'Hello world'` ?
If yes, you will get the following error: ` wrong number of arguments (given 2, expected 3)`

lib/sgtn-client/api/translation.rb
Xiaochao8: Yes. You are right.

if locale is empty string, then it can't be set to SgtnClient::Config.configurations.default.

`2.7.4 :007 > locale = ''
 => "" 
2.7.4 :008 > locale ||= 'abc'
 => "" 
2.7.4 :009 > locale
 => "" 
`

lib/sgtn-client/util/locale-util.rb
```diff
@@ -10,6 +10,11 @@ module SgtnClient
                   }
 
     class LocaleUtil
+        def self.process_locale(locale=nil)
+            locale = locale.class == Hash ? locale[:locale] : locale
```
Xiaochao8, line 14: In what case `locale` maybe a Hash?

linr211: same question. how the caller know it's used as locale[:locale], why not locale[:language] and etc.?

jessiejuachon: Is this not a valid call: `SgtnClient::Translation.getString("JAVA", "helloworld", locale:nil)`? If not, then I could remove hash handling. Even simpler.

jessiejuachon: I have removed the hash case.

Xiaochao8 (dismissed):

lib/sgtn-client/util/locale-util.rb
```diff
@@ -10,6 +10,10 @@ module SgtnClient
                   }
 
     class LocaleUtil
+        def self.process_locale(locale=nil)
+            locale ||= SgtnClient::Config.configurations.default
```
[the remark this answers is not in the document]
jessiejuachon: I believe it is loaded from sgtnclient.yml:
```
# # the default language used to load local source bundle
  default_language: en
```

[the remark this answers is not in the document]
jessiejuachon: It is an open [issue](https://github.com/vmware/singleton/issues/1478). Currently `SgtnClient::Config.configurations.default`  is being initialized [here](https://github.com/vmware/singleton/blob/6289d12579de534d6ffccd0205f46fe011d076cc/lib/sgtn-client/api/source.rb#L44). However, the value should have been configured configured in sgtnclient.yml.

Xiaochao8 (approved):

linr211 (approved):

Issue #1488, opened 2022-02-09
[BUG] [Ruby Client]Can't return expired message immediately when cache is expired.


lyiyu66:
**Describe the bug**
Can't return expired message immediately when cache is expired. 

**To Reproduce**
1. Set the client expired time to 60s.
2. Execute code as below: the requested component/key/locale is supported and existent.
```
@Result_localeTranslation = SgtnClient::Translation.getString("JAVA", "helloworld","fr")
puts "@Result: #@Result_localeTranslation"

#set the service is down
sleep 70

@Result_localeTranslation = SgtnClient::Translation.getString("JAVA", "helloworld","fr")
puts "@Result: #@Result_localeTranslation"
@Result_localeTranslation = SgtnClient::Translation.getString("JAVA", "helloworld","fr")
puts "@Result: #@Result_localeTranslation"

```
3. Check the output:
```
@Result: Bonjour le monde
@Result: s_Hello world
@Result: s_Hello world
```

**Expected behavior**
Simultaneously update the cache and return expired message to mitigate performance issues.
@Result: Bonjour le monde --- the 1st time, get the value from service
@Result: Bonjour le monde --- return the expired message
@Result: s_Hello world --- return the value of updated cache that was back from async thread in the 2nd request.

PR #1510, opened 2022-02-15
[Ruby Client]The 1st request after expiration can’t return expired message immediately. #1488

linr211:


tigershi (approved):

Xiaochao8:
will fix in 1579

PR #1496, opened 2022-02-10
Fixing issue 1479 - [REQUIREMENT] If the locale is of type Symbol, convert it first to String

jessiejuachon:

PR #1521, opened 2022-02-21
update the smoke test code beacuse pr 1495 and 1496

dleng-vmware:
1, Change the expect result, when locale is nil, not support language or int number, the get string method return default value rather than throw a error --- PR:https://github.com/vmware/singleton/pull/1495
2, If the locale is of type Symbol, convert it first to String, such as: :fr to "fr" ---PR:https://github.com/vmware/singleton/pull/1496

### DIFF
--- a/autotest/client/ruby/spec/SimpleT_spec .rb
+++ b/autotest/client/ruby/spec/SimpleT_spec .rb
@@ -46,7 +46,7 @@ describe "Translation test" do
         end
 
         it "Get a string's translation and locale not exist in server" do
-            RequestStore.store[:locale] = 'da'
+            RequestStore.store[:locale] = "da"
             RequestStore.store[:component] = 'about'
             #SgtnClient::Source.loadBundles("en")
             expect(SgtnClient::T.s("about.message")).to eq("fall back about")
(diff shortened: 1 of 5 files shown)


autotest/client/ruby/spec/SimpleT_spec .rb
```diff
@@ -46,7 +46,7 @@
         end
 
         it "Get a string's translation and locale not exist in server" do
-            RequestStore.store[:locale] = 'da'
+            RequestStore.store[:locale] = "da"
```
surite, line 49: any difference?

autotest/client/ruby/spec/Translation_offline_spec.rb
```diff
@@ -25,9 +25,9 @@
             expect(SgtnClient::Translation.getString("about", "about.message", "en")).to eq("fall back about")
         end
 
-        it "Get a string's translation and locale is fr-FR" do
+        it "Get a string's translation and locale is fr" do
             #SgtnClient::Source.loadBundles("de")
-            expect(SgtnClient::Translation.getString("about", "about.message", "fr-FR")).to eq("test fr offline key")
+            expect(SgtnClient::Translation.getString("about", "about.message", :fr)).to eq("test fr offline key")
```
surite, line 30: :fr is correct instead of "fr"?

autotest/client/ruby/spec/Translation_spec.rb
```diff
@@ -21,9 +21,9 @@
             expect(SgtnClient::Translation.getString("about", "about.message", "en")).to eq("fall back about")
         end
 
-        it "Get a string's translation and locale is fr-FR" do
+        it "Get a string's translation and locale is fr" do
             #SgtnClient::Source.loadBundles("de")
-            expect(SgtnClient::Translation.getString("about", "about.message", "fr-FR")).to eq("test fr key")
+            expect(SgtnClient::Translation.getString("about", "about.message", :fr)).to eq("test fr key")
```
surite, line 26: same question here about :fr

autotest/client/ruby/spec/Translation_offline_spec.rb
dleng-vmware: yes

autotest/client/ruby/spec/SimpleT_spec .rb
dleng-vmware: No difference, they are the same

autotest/client/ruby/spec/Translation_spec.rb
dleng-vmware: :fr is correct instead of "fr"

surite (approved):

PR #1556, opened 2022-02-28
change the l10n  the schedule strategy

tigershi:


g11n-ws/vip-manager-l10n/src/main/java/com/vmware/l10n/crons/SourceSendGRMCron.java
```diff
@@ -19,15 +19,15 @@
...
+	private final static long THREESECOND = 3000;
 	@Autowired
 	private SyncGrmSourceService syncGrmSourceService;
 	
 	
 	/**
 	 * Synchronize the updated source to remote GRM server
 	 */
-	@Scheduled(cron = "${sync.source.schedule.cron}")
+	@Scheduled(fixedDelay = THREESECOND)
```
linr211, line 30: why here it's @Scheduled(fixedDelay = THREESECOND), but in 'g11n-ws/vip-manager-l10n/src/main/java/com/vmware/l10n/crons/SourceLocalBundleCron.java' it's @Scheduled(cron = "${sync.source.schedule.cron}")?

tigershi: the tester requested that send to GRM source immediately after source merge to local bundle. there configure scanning the GRM source directory  every three second

PR #1559, opened 2022-03-02
[Ruby client] source comparison

jessiejuachon:


Xiaochao8:
close this because it's been implemented in another PR.

PR #1568, opened 2022-03-07
fix issue: throw expt when locale is nil

linr211:
fix issue: throw expt when locale is nil

### DIFF
--- a/lib/sgtn-client/api/source.rb
+++ b/lib/sgtn-client/api/source.rb
@@ -15,7 +15,7 @@ module SgtnClient
         else
           SgtnClient.logger.debug "[Source][getSource]getting sources from cache with key: " + cache_key
         end
-        s = items.nil?? nil : items[locale][key]
+        s = (items.nil? || items[locale].nil?)? nil : items[locale][key]  
         if items.nil? || s.nil?
           SgtnClient.logger.debug "[Source][getSource]source not found, return key: " + key
           #return key
--- a/lib/sgtn-client/api/source.rb
+++ b/lib/sgtn-client/api/source.rb
@@ -1,3 +1,7 @@
+# 
+#  Copyright 2019-2022 VMware, Inc.
+#  SPDX-License-Identifier: EPL-2.0
+#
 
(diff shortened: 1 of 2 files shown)


lib/sgtn-client/api/source.rb
```diff
@@ -15,7 +19,7 @@ def self.getSource(component, key, locale)
         else
           SgtnClient.logger.debug "[Source][getSource]getting sources from cache with key: " + cache_key
         end
-        s = items.nil?? nil : items[locale][key]
+        s = (items.nil? || items[locale].nil?)? nil : items[locale][key]
```
jessiejuachon, line 22: This is not correct. Per spec, if source bundle is not found, then it should be loaded right there and then.
Also, in CloudHealth, loadBundles had been invoked at app start up.

jessiejuachon (changes requested):

lib/sgtn-client/api/source.rb
linr211: I don't understand the reason, maybe Xiaochao could help explain it. However, from the log, it's this line throw the exception.

jessiejuachon: After adding this, what would the UI display then if no source strings is found?

linr211: It's fixing issues report in the logs, it's saying the array is nil before using it:
undefined method `[]' for nil:NilClass
["/home/cp-web/cloudpercept/shared/bundle/ruby/2.5.0/gems/singleton-client-0.7.4/lib/sgtn-client/api/source.rb:18:in `getSource'"

Xiaochao8 (dismissed):

lib/sgtn-client/api/source.rb
jessiejuachon: The array should not be nil in the first place.

jessiejuachon (dismissed):

Xiaochao8 (approved):

tigershi (approved):

PR #1577, opened 2022-03-08
[Ruby Client] Refactor

jessiejuachon:


lib/sgtn-client/api/source.rb
```diff
@@ -68,4 +38,4 @@ def self.getBundle(component, locale)
 
   end
 
-end
\ No newline at end of file
+end
```
jessiejuachon, line 41: Add bundle['messages'] = bundle.delete locale after line 32
```suggestion
end
```

lib/sgtn-client/api/translation.rb
```diff
@@ -137,12 +124,84 @@ def self.load_s(component, locale)
...
+        end
+        return thread
+      end
+    end
+
+    @@refresh_threads_lock = Mutex.new
+    @@refresh_threads = {}
+    def self.single_refresh(component, locale)
+      cache_key = SgtnClient::CacheUtil.get_cachekey(component, locale)
+      @@refresh_threads_lock.synchronize do
```
jessiejuachon, line 188: Only one thread in entire app can access this code block at a time. For better performance, it should be on1 thread per cached bundle.

Xiaochao8: This is necessary because `@@refresh_threads` is read at the beginning of this block and is assigned at the end.
No one statement can be moved out of this block.
This block is fast actually because it only starts a thread.

> For better performance, it should be 1 thread per cached bundle.
This is true bacause `@@refresh_threads` is a hash, each cached budle has an entry in it.

lib/sgtn-client/api/translation.rb
```diff
@@ -131,9 +130,11 @@ def self.compare_source(translation_bundle, old_source_bundle, source_bundle)
       translation_messages = translation_bundle['messages']
       translation_bundle['messages'] = new_translation_messages = {}
       source_bundle['messages'].each do |key, value|
-        new_translation_messages[key] = if old_source_messages[key] == value
-                                          translation_messages[key] || value
+        translation = translation_messages[key]
```
jessiejuachon, line 133: Also, we don't want to repetitively return an empty translation bundle in line 127.  I'd like to cache the fallback information once and use it until it expires.

lib/sgtn-client/api/translation.rb
```diff
@@ -158,5 +159,9 @@ def self.refresh_cache(component, locale)
       cache_key = SgtnClient::CacheUtil.get_cachekey(component, locale)
       @refresh_cache_operator.operate(cache_key, component, locale)
     end
+
+    String.class_eval <<-LOCALE, __FILE__, __LINE__ + 1
```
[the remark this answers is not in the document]
jessiejuachon: I get the idea, putting this info in the hash item itself instead of an an additional "fallback_keys" in the cache entry looks simpler. Any other ideas though? I don't want to modify Strings class.

lib/sgtn-client/api/translation.rb
jessiejuachon: This way, the approach will be consistent: no more post-processing after caching regardless if translation is missing for only one key or for the whole bundle. We also avoid multiple requests for the same missing bundle.

Xiaochao8:
Use another PR instead.

Issue #1558, opened 2022-03-02
 [BUG] [Ruby Client] (getStrings API)When requesting translation, client should return source if source has been changed. 


linr211:
For getStrings API, when requesting translation, client should return source if source has been changed.

lyiyu66:
Checked with build 0.7.0.36 with commit 09f7be8d62b86f6ae5c891a3c7d7e175059fa313, the issue is fixed.
Case1:
When some of keys in a component has been updated, requesting non-EN supported locale(example: ja), getStrings and get_translations() api return:
ja translation(for the source is same as messages_en.json) + source(en) string(for the source is not same as messages_en.json)
Case2:
New key/value pairs are added to source file, requesting non-EN supported locale(example: ja), getStrings and get_translations() api return:
ja translation(for the source is same as messages_en.json) + source(en) string(for the source is not same as messages_en.json if has) + new added source(en) string

Issue #1524, opened 2022-02-22
[BUG] [Ruby Client]Updated source string isn't displayed on UI. 


Xiaochao8:
This [PR](https://github.com/CloudHealth/cloudpercept/pull/21788) was deployed to labs and apps but the column name isn’t changing. 
When using It was working on local sandbox after changing the source.yml file. Do you know what could be going wrong in prod?

This is because Old English translation is used instead of source. We should figure out a way to reflect latest changes in source.

Issue #1485, opened 2022-02-09
[BUG] [Ruby Client]Redundant recheck in log file when requesting unsupported locale translation with getStrings().


lyiyu66:
**Describe the bug**
commit id:d95aed9ba515bfc45b0c5edd22b472d8cf2eec28
Redundant recheck in log file when requesting unsupported locale translation with getStrings().
Same issue in getString().

**To Reproduce**
1. Component "JAVA" is existing, locale "abc" isn't supported.
```
@Result_localeTranslation = SgtnClient::Translation.getStrings("JAVA", 'abc')
puts "@Result: #@Result_localeTranslation"
```
2. Check the log in stgnclient_d.log file. 
The process in red rectangle is not necessary since we already get the requested data in line 18: "Put cache for key 'test_4.8.1_JAVA_abc' with expired time at'2022-02-09 11:52:12 +0800" and it is not expired. Why we still need to recheck the data about "test_4.8.1_JAVA_default".
![image](https://user-images.githubusercontent.com/51813046/153114359-f9caf3e0-ef9f-45b3-9dd0-58710ebb0f09.png)

**Expected behavior**
No need the redundant process, just use the un-expired data directly.

lyiyu66:
Checked with build 0.7.0.36 with commit 09f7be8d62b86f6ae5c891a3c7d7e175059fa313, the log is changed since refactory and the redundant info has been removed.

Issue #1483, opened 2022-02-08
[BUG] [Ruby Client]"Source" is returned as locale value when requesting non-supported locale or "en" locale with getStrings().


lyiyu66:
**Describe the bug**
"Source" is returned as locale value when requesting non-supported locale or "en" locale with getStrings().

**To Reproduce**
1. Set source file is default.yml
2. Execute the code as below:
```
SgtnClient.load("./config/sgtnclient.yml", "test")
# load english bundle for fallback
SgtnClient::Source.loadBundles("default")
@Result_localeTranslation = SgtnClient::Translation.getStrings("JAVA", 'en')
puts "@Result: #@Result_localeTranslation"
```
3. See the behavior
`@Result: {"component"=>"JAVA", "messages"=>{"helloworld"=>"s_Hello world", "pluralsstr1"=>"%<{ \"cat_count\": { \"one\": \"s_there is one cat\", \"other\": \"s_there are %{cat_count} cats\" } }> in the room", "pluralsstr2"=>"%<{ \"cat_count\": { \"one\": \"s_there is one cat\", \"other\": \"s_there are %{cat_count} cats\" } }> in the cat_count room", "arg.usage"=>"%2$s, s_welcome login %1$s!", "arg.usage.1"=>"{A2}, s_welcome login {A1}!", "pluralsstrArg"=>"%%%%<{ \"cat_count\": { \"one\": \"s_there is one cat in %2$s and %1$s room\", \"other\": \"s_there are %%%%{cat_count} cats in %2$s and %1$s room\" } }>", "pluralsstrNamedArg"=>"%<{ \"cat_count\": { \"one\": \"s_there is one cat in {A2} and {A1} room\", \"other\": \"s_there are %{cat_count} cats in {A2} and {A1} room\" } }>", "helloworld_source"=>"s_Hello world_in source"}, "locale"=>"source"}`

**Expected behavior**
Should return the source locale "en" as value in this case:
`"locale"=>"en"`

lyiyu66:
Checked with build 0.7.0.36 with commit 09f7be8d62b86f6ae5c891a3c7d7e175059fa313, the issue is fixed with below 2 interfaces: getStrings() and get_translations()

PR #1579, opened 2022-03-09
allow one same request to server at the same time; add source comparison; revise locale utility

### LINKED ISSUE #1576: [REQUIREMENT] [Ruby] Revisit design and class diagrams, then refactor code
**Is your feature request related to a problem? Please describe.**

- [x] It is unclear what design patterns are used, if any, in the library. 
- [x] Some terminologies are inconsistent (e.g. "source" vs "default" vs "fallback")
- [x] Some code logic implementations are duplicated / complex.
- [x] Mark internal methods as private
- [x] Add source comparison
- [x] Only one same request(component-locale combination) is allowed at the same time
- [x] Config class
	-  It has many unnecessary code and needs to refractor.
	- Add necessary configuration items and remove uncessary configuration items.
- [x] Refractor cache module to use lock to make sure concurrency safe.
- [x] Refractor logger
- [x] Define new interfaces to make client easy-to-use.

**Describe the solution you'd like**
A clear and concise product design that may include class diagrams and and/or design patterns used.
Multiple small refactor-related PRs.

Xiaochao8:
Fixed #1576 #1524 #1381 #1362 #1293 #1558 #1488 #1485 #1483 #1460


lib/sgtn-client/common.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    # return new created object
+    def single_operation(id, *args)
+      @lock.synchronize do
+        obj = @hash[id]
+        @conditions.each do |con|
+          return obj unless con.call(id, obj, *args)
+        end
+        @hash[@id] = @creator.call(id, obj, *args)
```
jessiejuachon, line 23:
```suggestion
        @hash[id] = @creator.call(id, obj, *args)
```

jessiejuachon (changes requested):

lib/sgtn-client/common.rb
jessiejuachon: This error was not caught in unit test. It basically adds new thread as @hash[nil], which is also used for next calling thread. Could you update unit test to catch this error?

jessiejuachon: I think this will also solve the error you saw with get_cs on source bundle.

> https://github.com/vmware/singleton/runs/5508656880?check_suite_focus=true
> ThreadError:
>       ` Target thread must not be current thread`

lib/sgtn-client/api/translation.rb
```diff
@@ -1,158 +1,161 @@
...
+    end
+
+    def self.fetch(component, locale)
+      # source locale always uses source bundles
+      return Source.getBundle(component) if LocaleUtil.is_source_locale(locale)
+
+      translation_bundle_thread = Thread.new { load(component, locale) }
+      source_bundle_thread = Thread.new { Source.getBundle(component) }
+      old_source_bundle = load(component, LocaleUtil.get_source_locale)
+      source_bundle = source_bundle_thread.value
```
jessiejuachon, line 122: Not sure why get_cs was calling .value on current thread. 

I think this should now work.
```suggestion
      old_source_bundle = load(component, LocaleUtil.get_source_locale)
      source_bundle = get_cs(component, LocaleUtil.get_source_locale)
```

jessiejuachon (changes requested):

lib/sgtn-client/common.rb
Xiaochao8: Same error after change.

lib/sgtn-client/api/translation.rb
Xiaochao8: Don't wok actually.

spec/unit/locale_spec.rb
```diff
@@ -35,28 +35,55 @@
...
+    #   expect(SgtnClient::LocaleUtil.process_locale('en')).to eq 'en'
+    #   expect(SgtnClient::LocaleUtil.process_locale('de-DE')).to eq 'de-DE'
+    #   expect(SgtnClient::LocaleUtil.process_locale(:'de-DE')).to eq 'de-DE'
+    #   expect(SgtnClient::LocaleUtil.process_locale(:ja)).to eq 'ja'
+    #   expect(SgtnClient::LocaleUtil.process_locale(:'ja')).to eq 'ja'
+    # end
+
+    # it "process_locale_nil" do
+    #   expect(SgtnClient::LocaleUtil.process_locale(nil)).to eq SgtnClient::Config.configurations.default
+    # end
```
jessiejuachon, line 48: Remove these if not needed.

spec/unit/source_spec.rb
```diff
@@ -16,17 +14,17 @@
...
+    #   key="new_welcome"
+    #   str = SgtnClient::Source.getSource(@component, key, @locale)
+    #   expect(str).to eq 'New %2$s, welcome login %1$s!'
+    # end
+
+    # it "getSources" do
+    #   str = SgtnClient::Source.getSources(@component, @locale)
+    #   puts str[@locale]
+    #   expect(str).not_to eq be_nil
+    # end
```
jessiejuachon, line 27: Remove these if not needed. Add a test for loadBundle.

jessiejuachon (changes requested):

lib/sgtn-client/util/locale-util.rb
```diff
@@ -2,45 +2,49 @@
...
+      parts = locale.split('-')
+      if parts.size > 1
+        f = SgtnClient::DEFAULT_LOCALES.select { |e| e == parts[0] }
+        return f[0] unless f.empty?
+      end
+      locale
+    end
+
+    def self.get_source_locale
+      'en'
```
jessiejuachon, line 42: What if product's source locale is not 'en'?

jessiejuachon (changes requested):

jessiejuachon:
Let's get this merged and have small PRs moving forward.

deminyan (approved):

Issue #1533, opened 2022-02-22
[BUG] [Java Client]java.lang.NullPointerException pops up when component is "null" in getMessage() interface. 


lyiyu66:
**Describe the bug**
Exception pops up when component is "null" in getMessage() interface.
Get same exception "java.lang.NullPointerException" when getMessages() uses "null" component.

**To Reproduce**
1. Invoke getMessage() as below:
`String msg3 = tm.getMessage(<locale>, null, <existentKey>);`
2. Get the below exception:
![image](https://user-images.githubusercontent.com/51813046/155114874-f22daa24-0f70-4100-a8b2-87460a508b93.png)

**Expected behavior**
Should get an exception that clarify what the problem is.
Like: Failed to get any message for key: <key> of component <component>, requested locale: <locale>

lyiyu66:
Checked with commit 5debd59f1b4085b0784c3025b7f28206357c0964:
1. component=null, will pops up message "Component can't be empty!"
2. component=nonExistingComponentName, will pops up message "Failed to get any message for key: <requestedKey> of component <requestedComponentName>, requested locale: <requestedLocale>"

Issue #1515, opened 2022-02-16
[BUG] [Java Client]Exception pops up when locale is "null" in getMessage() interface.


lyiyu66:
**Describe the bug**
Exception pops up when locale is "null" in getMessage() interface.
Get same exception "java.lang.NullPointerException" when getMessages() uses "null" locale.

**To Reproduce**
1. Invoke getMessage() as below:
`String msg3 = tm.getMessage(null, <existentComponent>, <existentKey>);`
2. Get the below exception:
![image](https://user-images.githubusercontent.com/51813046/154222345-7dce2000-b5b8-4b11-b45e-684809993e60.png)

**Expected behavior**
Return fallback locale's translation: default locale -> source locale

lyiyu66:
Verified with commit 5debd59f1b4085b0784c3025b7f28206357c0964, it can return to fallback locale when requested locale is null.

PR #1583, opened 2022-03-09
fix bug 1515: locale null exception and 1533: component null exception

huihuiw01:
#1515: [BUG] [Java Client]Exception pops up when locale is "null" in getMessage() interface.
and
#1533: [BUG] [Java Client]java.lang.NullPointerException pops up when component is "null" in getMessage() interface.

### DIFF
--- a/src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
+++ b/src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
@@ -133,7 +133,7 @@ public class TranslationMessage implements Message {
         if(locale == null){
             throw new VIPJavaClientException(ConstantsMsg.LOCALE_CANNOT_NULL);
         }
-        if(component == null || "".equalsIgnoreCase(component)){
+        if(StringUtil.isEmpty(component)){
             throw new VIPJavaClientException(ConstantsMsg.COMPONENT_CANNOT_EMPTY);
         }
         ComponentService.TranslationsDTO msgsItemDTO = getMessages(locale, component, true);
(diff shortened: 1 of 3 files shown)


src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -130,7 +130,12 @@ private String getMessageWithArgs(String resourceBundle, final Locale locale, fi
                 return FormatUtils.formatMsg(source, LocaleUtility.getSourceLocale(), args);
             }
         }
-
+        if(locale == null){
```
Xiaochao8, line 133: Don't need `"".equalsIgnoreCase(locale)`?

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -130,7 +130,12 @@ private String getMessageWithArgs(String resourceBundle, final Locale locale, fi
...
+            throw new VIPJavaClientException(ConstantsMsg.LOCALE_CANNOT_NULL);
+        }
+        if(component == null || "".equalsIgnoreCase(component)){
```
Xiaochao8, line 136: I remember there is a method StringUtil.Empty, isn't it?

tigershi, line 136: the component name is case sensitive

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
huihuiw01: Yes, because it's a Locale type object, not a String type object.

src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
huihuiw01: When component's value is empty, it doesn't matter to use case sensitive or insensitive method when compare.

Xiaochao8 (approved):

tigershi (approved):

PR #1588, opened 2022-03-11
Add # in key validate

tigershi:


linr211 (approved):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/utils/RegExpValidatorUtils.java
```diff
@@ -66,6 +66,14 @@ public static boolean IsLetterAndNumberAndValidchar(String str) {
 		String regex = "^[A-Za-z0-9_\\-\\.]+$";
 		return match(regex, str);
 	}
+
```
huihuiw01, line 69: Should the old method 'IsLetterAndNumberAndValidchar' be removed?

huihuiw01 (approved):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/utils/RegExpValidatorUtils.java
tigershi: other method still use it

PR #1593, opened 2022-03-11
Fix #1575 File read does not support thread concurrency

### LINKED ISSUE #1575: [BUG] [Ruby Client] File read does not support thread concurrency
**Describe the bug**
File read for source bundle is currently inside a synchronized block. This means only one thread can access it at a time, affecting performance.

**To Reproduce**
1. Use the ruby client lib in a sample app.
2. Launch multiple(e.g. 1,000) threads that will load a source bundle at the same time. Use a bundle that has a large data set to observe performance.
3. Notice that the threads wait for their turn and so affecting performance.
4. Do the same for any offline bundle and see the same performance hit.

**Expected behavior**
Multiple threads should be able to read the source bundle at the same time.

jessiejuachon:


lib/sgtn-client/util/file-util.rb
```diff
@@ -7,31 +7,23 @@
...
 
       def self.read_yml(file_name)
-        SgtnClient.logger.debug "[FileUtil]read yml file from: " + file_name
-        @mutex.synchronize do
-          erb = ERB.new(File.read(file_name))
-          erb.filename = file_name
-          YAML.load(erb.result)
+        File.open(file_name) do |file|
+          sources = YAML::load(file)
         end
```
Xiaochao8, line 25: Looks this method is unncessary.
```suggestion
        YAML::load(File.read(file_name))
```

lib/sgtn-client/util/file-util.rb
```diff
@@ -7,31 +7,23 @@
...
-          end
-          return data_hash
+        data_hash = nil
+        begin
+          file = File.read(bundlepath)
+          data_hash = MultiJson.load(file)
+        rescue => exception
+          SgtnClient.logger.error exception.message
         end
+        return data_hash
```
Xiaochao8, line 19:
```suggestion
        begin
          return MultiJson.load(File.read(bundlepath))
        rescue => exception
          SgtnClient.logger.error exception.message
        end
        nil
```

Xiaochao8 (approved):

Xiaochao8 (approved):

deminyan (approved):

PR #1603, opened 2022-03-15
[Ruby Client] Fixing issues 1602 and 1604

jessiejuachon:


lib/sgtn-client/util/cache-util.rb
```diff
@@ -8,25 +8,25 @@ module Core
 
   class CacheUtil
     def self.get_cache(cache_key)
-      expired, items = SgtnClient::Core::Cache.get(cache_key)
-      SgtnClient.logger.debug "[CacheUtil]get cache with key #{cache_key}, expired #{expired}"
-      [expired, items]
+      cache_item = SgtnClient::Core::Cache.get(cache_key)
+      SgtnClient.logger.debug "[CacheUtil]get cache with key #{cache_key}"
+      cache_item
```
Xiaochao8, line 13:
```suggestion
      SgtnClient.logger.debug "[CacheUtil]get cache with key #{cache_key}"
      SgtnClient::Core::Cache.get(cache_key)
```

lib/sgtn-client/api/translation.rb
```diff
@@ -50,32 +42,26 @@ def self.getString_f(component, key, args, locale, *optionals)
...
       SgtnClient.logger.debug "[Translation][getStrings]component=#{component}, locale=#{locale}"
       locale = SgtnClient::LocaleUtil.get_best_locale(locale)
-      items = get_cs(component, locale)
-      if (items.nil? || items['messages'].nil?) && !LocaleUtil.is_source_locale(locale)
-        items = get_cs(component, LocaleUtil.get_source_locale)
-      end
-
-      items
+      items = get_cs(component, locale)&.dig(:items)
+      items = items&.dig('messages').nil? && !LocaleUtil.is_source_locale(locale) ? get_cs(component, LocaleUtil.get_source_locale)&.dig(:items) : items
```
Xiaochao8: This looks more clear.
```suggestion
      if items&.dig('messages').nil? && !LocaleUtil.is_source_locale(locale)
        get_cs(component, LocaleUtil.get_source_locale)&.dig(:items)
      else
        items
      end
```

lib/sgtn-client/api/translation.rb
```diff
@@ -12,24 +12,16 @@ module Core
   module Translation
     def self.getString(component, key, locale)
       SgtnClient.logger.debug "[Translation.getString]component: #{component}, key: #{key}, locale: #{locale}"
-      str = getTranslation(component, key, locale)
-      if str.nil? && !LocaleUtil.is_source_locale(locale)
-        str = getTranslation(component, key, LocaleUtil.get_source_locale)
-      end
-      str
+      getTranslation(component, key, locale)
```
Xiaochao8, line 15: Isn't `getTranslation` possible return `nil`?

jessiejuachon: It won't anymore because of the following line in fetch:` return source_bundle, nil, LocaleUtil.get_source_locale if translation_bundle.nil? `

This means we always write something in cache. If translation fetch for target locale fails, we write a cache_item that has a reference to the source bundle (because it has fallen back to the source bundle). This avoids the many requests for a translation bundle that is not yet ready, or for when fetch fails for any other reason (e.g. VIP service is down).

Take note that we won't duplicate the source bundle in memory; it only saves a reference to the already existing source bundle in memory.

Xiaochao8:
This has been fixed.

PR #1612, opened 2022-03-17
Fixing 1602

jessiejuachon:


lib/sgtn-client/util/string-util.rb
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022 VMware, Inc.
+# SPDX-License-Identifier: EPL-2.0
+
+module SgtnClient
+  class StringUtil < String
+    def initialize(str, locale)
+      super(str)
+      @locale = locale
+    end
+    attr_accessor :locale
```
Xiaochao8, line 10: Seems no need to change the locale.
```suggestion
    attr_reader :locale
```

Xiaochao8 (approved):

deminyan (approved):

PR #1614, opened 2022-03-18
cache code refactor

jessiejuachon:


spec/unit/inconfig_spec.rb
```diff
@@ -13,7 +13,7 @@
     end
 
     it "validate_configuration" do
-      SgtnClient.load("./spec/config/sgtnclient-invalidate.yml", "test", './sgtnclient_config.log')
```
Xiaochao8, line 16: Why comment this line?

Xiaochao8 (approved):

spec/unit/inconfig_spec.rb
jessiejuachon: This loads an invalid config file, which gets consumed by next set of tests.

deminyan (approved):

PR #1620, opened 2022-03-21
write creation.json before translation bundle files 

tigershi:


jin-jasper (dismissed):

linr211:
I don't see the difference after change, could you explain a bit why to have this change?

jin-jasper (approved):

tigershi:
the script need to update the create.json before update the bundle content

tigershi:
don't see the difference after change in business

PR #1621, opened 2022-03-21
Refractor to use one string interface

Xiaochao8:


tigershi (approved):

tigershi:
Not pass the unit test

lib/sgtn-client/api/translation.rb
```diff
@@ -11,42 +11,39 @@ module Core
...
       str = getTranslation(component, key, locale)
-      if str.nil? && !LocaleUtil.is_source_locale(locale)
-        str = getTranslation(component, key, LocaleUtil.get_source_locale)
+      if str
+        locale = str.locale if str.is_a?(SgtnClient::StringUtil)
+      elsif !LocaleUtil.is_source_locale(locale)
+        locale = LocaleUtil.get_source_locale
+        str = getTranslation(component, key, locale)
+      end
+      return if str.nil?
```
jessiejuachon, line 24: Will this return nil?

lib/sgtn-client/api/translation.rb
```diff
@@ -11,42 +11,39 @@ module Core
...
+        str.localize(locale) % kwargs
+      elsif !args.empty?
+        str.localize(locale) % args
+      else
+        str
       end
-      str
     end
 
     def self.getString_p(component, key, plural_args, locale)
```
jessiejuachon, line 35: Is a separate getString_p necessary? Why not call getString directly? So instead of
`SgtnClient::Translation.getString_p("JAVA", "plural_key", { :cat_count => 1 }, "zh-Hans")`, 
call `SgtnClient::Translation.getString("JAVA", "plural_key", "zh-Hans", { :cat_count => 1 })`

lib/sgtn-client/api/translation.rb
```diff
@@ -11,42 +11,39 @@ module Core
...
-          str.to_plural_s(LocaleUtil.get_source_locale, plural_args) if str
-        end
-      else
-        locale = str.locale if str.is_a?(SgtnClient::StringUtil)
-        str.to_plural_s(locale, plural_args)
-      end
+      getString(component, key, locale, **plural_args)
     end
 
     def self.getString_f(component, key, args, locale, *optionals)
```
jessiejuachon, line 40: Is a separate getString_f necessary? Why not call getString directly, passing the intended arguments?

jessiejuachon (changes requested):

lib/sgtn-client/api/translation.rb
Xiaochao8: This is for compatibility. Can consider to remove them.

lib/sgtn-client/api/translation.rb
Xiaochao8: This is for compatibility. Can consider to remove them.

lib/sgtn-client/api/translation.rb
```diff
@@ -61,9 +58,7 @@ def self.getStrings(component, locale)
     end
 
     def self.getTranslation(component, key, locale)
```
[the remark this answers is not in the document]
Xiaochao8: Agree

lib/sgtn-client/api/translation.rb
jessiejuachon: If only for compatibility with existing code in CHT, we can mark this as deprecated and suggest to use get_string. We can do this in another PR if you like.

lib/sgtn-client/api/translation.rb
jessiejuachon: Same comment as above.

lib/sgtn-client/api/translation.rb
```diff
@@ -11,42 +11,39 @@ module Core
   end
 
   module Translation
-    def self.getString(component, key, locale)
-      SgtnClient.logger.debug "[Translation.getString]component: #{component}, key: #{key}, locale: #{locale}"
+    def self.getString(component, key, locale, *args, **kwargs)
```
jessiejuachon, line 14:
```suggestion
    # Retrieves translation of a message
    #
    # @example
    #
    #    translation1 = SgtnClient::Translation.getString("some_component", "some_key", "some_locale", { :count => 1 })
    #    translation2 = SgtnClient::Translation.getString("some_component", "some_key_2")
    #
    # @param [String] component Localization component
    # @param [String] key Key to identify the message
    # @param [String or Symbol] locale Translation locale
    # @param args Optional arguments passed from input
    # @param [Hash<Symbol, Integer>] kwargs Optional hash with symbol keys and integer values for plural arguments
    # 
    # @return [String] Translation
    def self.getString(component, key, locale, *args, **kwargs)
```

lib/sgtn-client/api/translation.rb
```diff
@@ -11,42 +11,39 @@ module Core
...
-      if str.nil?
-        unless LocaleUtil.is_source_locale(locale)
-          str = getTranslation(component, key, LocaleUtil.get_source_locale)
-          str.to_plural_s(LocaleUtil.get_source_locale, plural_args) if str
-        end
-      else
-        locale = str.locale if str.is_a?(SgtnClient::StringUtil)
-        str.to_plural_s(locale, plural_args)
-      end
+      getString(component, key, locale, **plural_args)
```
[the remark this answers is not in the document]
jessiejuachon: > Is to_plural_s in localized_str.rb used? If not, please remove the whole file and the related tests in plural_spec.rb. Remove it from README too.

Should this be removed?

jessiejuachon (changes requested):

lib/sgtn-client/api/translation.rb
Xiaochao8: yes.

Xiaochao8:
design is changed. close this.

PR #1677, opened 2022-04-27
launch only one thread when cache is expired

Xiaochao8:


lib/sgtn-client/sgtn-client.rb
```diff
@@ -4,6 +4,9 @@
 module SgtnClient
       LOGFILE_SHIFT_AGE = 4
 
+      module Common
+            autoload :BundleID,     "sgtn-client/common/data"
+      end
```
jessiejuachon, line 9: This uses only the locale and component. Should we consider adding product name and version ID as well if in case we want to support being able to load translation for multiple VIP "products" in one application?

jessiejuachon:
Please take a look at the following failing local tests:
Screen Shot 2022-05-24 at 18 38 04

Xiaochao8:
> 
Please use another higher ruby version such as 2.7 to run tests. Ruby 2.5.5 has a bug possibly.

Xiaochao8:
> Please take a look at the following failing local tests: Screen Shot 2022-05-24 at 18 38 04

Please use another higher ruby version such as 2.7 to run tests. Ruby 2.5.5 has a bug possibly.

lib/sgtn-client/sgtn-client.rb
Xiaochao8: Currently our ruby client doesn't support multiple products. Supporting it will need a big change.
Our interfaces need to add product and version as arguments.

jessiejuachon: Okay, will think more about it later on.

jessiejuachon:
> > Please take a look at the following failing local tests: Screen Shot 2022-05-24 at 18 38 04
> 
> Please use another higher ruby version such as 2.7 to run tests. Ruby 2.5.5 has a bug possibly.

Wouldn't this affect CloudHealth which is using 2.5.5? Tests should pass in 2.5.5 as well, right?

Xiaochao8:
> > > Please take a look at the following failing local tests: Screen Shot 2022-05-24 at 18 38 04
> > 
> > 
> > Please use another higher ruby version such as 2.7 to run tests. Ruby 2.5.5 has a bug possibly.
> 
> Wouldn't this affect CloudHealth which is using 2.5.5? Or is this in tests only? Could you post a reference to the 2.5.5 bug please?

No. The code which has problems in 2.5.5 is only used in tests. See here: https://github.com/Xiaochao8/singleton/blob/db0d6127f5370a6ef2c5ce49768724139d7700d8/spec/loader/single_loader_spec.rb#L29

jessiejuachon (dismissed):

deminyan (approved):

jessiejuachon (approved):

PR #1745, opened 2022-06-20
set only one configuration item once

Xiaochao8:


lib/sgtn-client/util/validate-util.rb
```diff
@@ -6,31 +6,31 @@ module SgtnClient
...
         end
         
-        disable_cache = Config.configurations[env]["disable_cache"]
+        disable_cache = SgtnClient.config.disable_cache
         if disable_cache != nil && disable_cache != false && disable_cache != true
           messages = messages +  "Configuration[disable_cache] has to be a 'true' or 'false'!\n"
         end
         
-        if messages != "\n"
+        if messages != ""
```
jessiejuachon, line 33: Could you change "messages" to something more descriptive like "error_messages"? Also, could you make this a list/array instead of one string and pass to MissingConfig?

This can be in a separate PR if you prefer.

jessiejuachon (changes requested):

lib/sgtn-client/util/validate-util.rb
Xiaochao8: Yes. Let's change it in another PR.

Xiaochao8:
> Could you fix the codacy issues and smoke test failures?

codacity issues are fixed.
Devops is fixing an issue running smoke test.

jessiejuachon (approved):

PR #1748, opened 2022-06-21
G11n cldr pattern add source jar and java doc jar

tigershi:


src/test/java/com/vmware/test/i18n/I18nUtilTest.java
```diff
@@ -25,7 +25,7 @@ public class I18nUtilTest {
 	final String testURL = "https://github.com/unicode-cldr/cldr-core/archive/32.0.0.zip";
 	final String[] catesArr = { "dates", "numbers", "plurals", "measurements"};
 
-	@Before
```
Xiaochao8, line 28: Why commented?

Xiaochao8 (approved):

src/test/java/com/vmware/test/i18n/I18nUtilTest.java
tigershi: because this is test the  download 32.0.0 cldr data. we don't need to test it

Xiaochao8 (approved):

linr211 (approved):

PR #1786, opened 2022-07-12
upgrade springboot to 2.5.14

tigershi:


tigershi:
It doesn't meet the requirement

PR #1790, opened 2022-07-14
G11n java client javadoc format

tigershi:


src/main/java/com/vmware/vipclient/i18n/base/instances/TranslationMessage.java
```diff
@@ -21,8 +21,8 @@
 /**
  * This class provide some APIs to get translation from VIP service in
  * string-based, component-based level.
- *
- * For string-based level APIs, some need the source string as parameter, some don't, instead the source string is put to the resource bundle.
+ * <pre>
+ * <p>For string-based level APIs, some need the source string as parameter, some don't, instead the source string is put to the resource bundle.
```
Xiaochao8, line 25: What does <p> mean?

tigershi: format java doc

Xiaochao8: Is `</p>` missed?

linr211:
what's the JIRA  or github ticket? I don't understand why needs to include the java doc and source files into the jar build.

linr211 (approved):

PR #1884, opened 2022-09-19
Update codecov.yml to fit configure format change

junw-33455432:


nannany00 (approved):

codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+  precision: 2        
+  round: down         
+  range: 20...80     
+  status:
+    project:                   
+      default:                 
+        enabled: yes           
+        target: 80%                 
+        if_no_uploads: error   
+        if_not_found: error
```
surite, line 11: option  should be one of 'success' and 'failure', refer to https://docs.codecov.com/docs/commit-status#if_not_found

codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+        target: 80%                 
+        if_no_uploads: error   
+        if_not_found: error  
+        if_ci_failed: error   
+    patch:
+      default:
+        enabled: true            
+        target: 80%                      
+        if_no_uploads: error     
+        if_not_found: error
```
surite, line 18: same with above

codecov.yml
```diff
@@ -0,0 +1,22 @@
+coverage:
+  precision: 2        
+  round: down         
+  range: 20...80
```
surite, line 4: if the target is 80%, why the green range is less than 80%?

codecov.yml
junw-33455432: Thanks for the comment, it has been updated.

codecov.yml
junw-33455432: Updated as above one.

surite (approved):

renligeng (approved):

renligeng (approved):

PR #1911, opened 2022-10-10
support pseudo feature

Xiaochao8:
Need to test I18n backend pseudo function


lib/sgtn/pseudo.rb
```diff
@@ -0,0 +1,34 @@
...
+    def initialize
+      @source_locale = SgtnClient::LocaleUtil.get_source_locale
+      @prefix = Sgtn.pseudo_prefix || '@@'
+      @suffix = Sgtn.pseudo_suffix || @prefix
+    end
+
+    def get_string!(key, component, locale)
+      if Sgtn.pseudo_mode
+        # source is always available, so actual_locale is same as @source_locale
+        translation, actual_locale = super(key, component, @source_locale)
```
linr211, line 17: for online mode, should set the query parameter 'pseudo' to be 'true' when request the pseudo translation from service?

spec/i18n_backend_spec.rb
```diff
@@ -45,4 +45,19 @@
     I18n.locale = en_locale
     expect(I18n.t(key)).to eq en_value
   end
+
+  it '#pseudo mode' do
+    Sgtn.pseudo_mode = true
```
linr211, line 50: When do integration project, can the product team's developer enable the pseudo feature on the product's backend by this codeline 'Sgtn.pseudo_mode = true' as well?

linr211: I just want to confirm the usage.

lib/sgtn/pseudo.rb
Xiaochao8: It's unnecessary, client will retrieve 'latest' locale and add pseudo tags. They are same.

spec/i18n_backend_spec.rb
Xiaochao8: Yes.

lib/sgtn/pseudo.rb
linr211: If so, is the pseudo tag same with service '#@'?

Xiaochao8: You can set tags to '#@' by configuration item `pseudo_prefix` and `pseudo_suffix`.

linr211: how to identify if the pseudo translation comes from online or offline? seems the tags are same for both modes.

Xiaochao8: It's configurable. You can set tags by mode.

linr211: Seems not. Is there an example?

 For this setting as example:
     _vip_server: https://g11n-vip-stg-1.eng.vmware.com:8090
     pseudo_mode: true
     pseudo_prefix: "@@"
     pseudo_suffix: "@@"_
If the locale is source_locale, it will get local source, so the pseudo is appended with '@@'(such as @@local_source@@); if the locale is not source_locale, it will get latest source from remote service, but the pseudo is appended with same tag '@@' (such as @@remote_latest@@).

Xiaochao8: In pseudo mode, client will return pseudo translation for all locales.

tigershi (approved):

spec/config/sgtnclient.yml
```diff
@@ -24,6 +24,9 @@ test: &default
   # # the lowest level to be logged, default value is INFO
   log_level: DEBUG
 
+  # # tag for pseudo translation, default value is '@@'
+  pseudo_tag: "@@"
```
linr211, line 28: no double quotes?

lib/singleton-client/pseudo.rb
```diff
@@ -0,0 +1,14 @@
...
+
+# Copyright 2022 VMware, Inc.
+# SPDX-License-Identifier: EPL-2.0
+
+module Sgtn
+  module Pseudo # :nodoc:
+    protected
+
+    def pickup_locale(locale, _component)
+      locale == PSEUDO_LOCALE ? PSEUDO_LOCALE : super
```
linr211, line 11: if locale=pseudo, will return pseudo translation?
I don't see how to trigger and get the pseudo from Singleton service, the mocked testing just show get from a file but not from a pseudo locale. Could you explain a bit how remote pseudo work?

Xiaochao8: > if locale=pseudo, will return pseudo translation?
Yes. It will

It is limited by current client design, client will get latest source and add pseudo tags on source strings. This should meet your fundamental pseudo requirements. As we've discussed before, it's hard to get pseudo from server to reflect source collection status in current design. I've tried that way and the logic was complex and some issues didn't find a solution. We can discuss more on this offline.

spec/config/sgtnclient.yml
Xiaochao8: Looks @@ has special meaning in YAML, the double quotes are necessary.

lib/singleton-client/pseudo.rb
linr211: i think it has to update the cache by another request to remote latest with pseudo locale, you can consider to implement this by another PR.

linr211 (approved):

PR #1935, opened 2022-11-04
fixed the componentlist API no error msg in S3

tigershi:


Xiaochao8 (changes requested):
IMO, it's acceptable to return 200 and empty array when no components.
Unnecessary to add extra code to deal with this.

linr211:
> IMO, it's acceptable to return 200 and empty array when no components. Unnecessary to add extra code to deal with this.

I agree, until there's real error when get the component list.

PR #1938, opened 2022-11-10
apply IAM role to AWS S3 account

Xiaochao8:
Other than applying IAM role to Aws account, this PR upgrades Aws S3 Golang interfaces from V1 to V2.


modules/translation/dao/s3bundle/s3client.go
```diff
@@ -0,0 +1,99 @@
...
+		RoleCredentials: roleOutput.Credentials,
+	}
+
+	writeClient(newInst)
+
+	return newInst
+}
+
+func readClient() RoleClient {
+	clientInstLock.RLock()
```
tigershi, line 88: why need add a lock in read

Xiaochao8: Here is the RWLock - **clientInstLock** - to protect read and write.

Xiaochao8: It's possible that multiple threads access the instance symultaneously.

jin-jasper (dismissed):

modules/translation/dao/s3bundle/s3client.go
tigershi: You have used the clientInstLock in write function and it is sync.RWMutex. And you use the other Lock in read function. The one source use the two locks. It means that read function and write function can concurrent execution

Xiaochao8: The one resource use the same lock `clientInstLock` which is a  `sync.RWMutex`. When reading the variable, use a read lock; When writting the variable, use write lock.
Are there any problems?

tigershi (approved):

modules/translation/dao/s3bundle/s3client.go
Xiaochao8: Resovled to use a signal variable.

tigershi (approved):

jin-jasper (approved):

PR #1968, opened 2022-11-25
Compatible with previous code which is able to read L2 bundle binded in pattern jar package to fix bug 1953 - Singleton service CI failure and Java Client offline mode failure

huihuiw01:


huihuiw01:
This PR is for testing CI problem, close it.

PR #2017, opened 2022-12-26
source should be collected successfully when csp token setting true

tigershi:


jin-jasper (approved):

Xiaochao8 (approved):

linr211:
It's not easy to figure out the new behavior from the code changes, where's the issue link/description about the expected source collection behavior on CSP?

jin-jasper (approved):

PR #2018, opened 2022-12-28
add l10n manager request log interceptor

tigershi:


Xiaochao8:
What issue are you fixing?

linr211:
Same question with Xiaochao that there's lack of background information for your change, so please try to decribe more details or paste the issue link

tigershi:
Because the l10n have been forwarded by gateway when source collection, we need some information from other team in Request Header to follow-up the API request process in l10n log.

g11n-ws/vip-manager-l10n/src/main/java/com/vmware/l10n/conf/RequestLogInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.Arrays;
+import java.util.List;
+
+public class RequestLogInterceptor extends HandlerInterceptorAdapter {
```
linr211, line 18: i think it's not necessary to create a new interceptor which can decrease the performance of request handling, so can you add the log message about request-id in CspAuthInterceptor?

Xiaochao8 (dismissed):

g11n-ws/vip-manager-l10n/src/main/java/com/vmware/l10n/conf/RequestLogInterceptor.java
tigershi: only can add the log in CollectSourceValidationInterceptor and have done!

huihuiw01 (approved):

linr211 (approved):

PR #2041, opened 2023-01-10
add base64 decode in l10n api

tigershi:


jin-jasper (approved):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/constants/ConstantsKeys.java
```diff
@@ -106,6 +106,7 @@ public class ConstantsKeys {
    public static final String SCOPE_FILTER_MATCH = "(^[a-zA-Z]+[,-_a-zA-Z0-9]+[a-zA-Z0-9]+$)|(^[\\^\\(]+[,-_a-zA-Z0-9]+[\\)]+$)";
 
    public final static Set<String> SOURCE_FORMAT_LIST = new HashSet<String>(Arrays.asList("STRING", "SVG", "MD", "HTML"));
+   public final static String SOURCE_FORMAT_BASE64 = "BASE64";
```
linr211, line 109: why not move "BASE64" to SOURCE_FORMAT_LIST as one of elements?

g11n-ws/vip-manager-l10n/src/main/java/com/vmware/l10n/conf/CollectSourceValidationInterceptor.java
```diff
@@ -166,13 +167,20 @@ private static void validateLocale(HttpServletRequest request)
...
+			if (SourceFormatUtils.isBase64Encode(sourceFormat)){
+				sourceFormat = SourceFormatUtils.formatSourceFormatStr(sourceFormat);
+				if (!StringUtils.isEmpty(sourceFormat) && !ConstantsKeys.SOURCE_FORMAT_LIST.contains(sourceFormat)) {
+					throw new ValidationException(ValidationMsg.SOURCEFORMAT_NOT_VALIDE);
+				}
+			}else{
+				if (!ConstantsKeys.SOURCE_FORMAT_LIST.contains(sourceFormat)) {
+					throw new ValidationException(ValidationMsg.SOURCEFORMAT_NOT_VALIDE);
+				}
+			}
```
linr211, line 183: if my above comment 'why not move "BASE64" to SOURCE_FORMAT_LIST as one of elements?' is true, i think these codes can be removed.

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/constants/ConstantsKeys.java
tigershi: because we need set sourceformat to (BASE64,STRING), (BASE64,SVG) ... when use the base64 encode in clients , if the BASE64 in SOURCE_FORMAT_LIST, we can't compatible this case

g11n-ws/vip-manager-l10n/src/main/java/com/vmware/l10n/conf/CollectSourceValidationInterceptor.java
tigershi: we need set sourceformat to (BASE64,STRING), (BASE64,SVG) ... when use the base64 encode in clients

linr211 (approved):

PR #2042, opened 2023-01-11
add multiple components interface

Xiaochao8:
Changes in this PR

- Get translation with `Get Product component API` rather than `Get Product translation API` which is used before.
- Add multiple components interface by calling single component interface parallelly.
- Some minor changes on Singleton server URLs.


tigershi (dismissed):

tigershi (approved):

translation_interface.go
```diff
@@ -18,4 +18,7 @@ type Translation interface {
 
 	// GetComponentMessages Get component messages
 	GetComponentMessages(name, version, locale, component string) (ComponentMsgs, error)
+
+	// GetComponentsMessages Get messages of multiple components
+	GetComponentsMessages(name, version string, locales, components []string) ([]ComponentMsgs, error)
```
linr211, line 23: is it necessary to get more than one locale's transaltions?

tigershi (approved):

translation_interface.go
Xiaochao8: Yes. Because the server may serve clients with different locales.

linr211 (approved):

PR #2242, opened 2023-04-10
The file application-bundle.properties is changed, so that the smoke …

Aleksbogdanov12:
…test for python client

can work properly. The python script update_config_files_with_abs_paths.py is used to modify all the testcases in the python client, so that the config files are using the absolute path. Github action does not work with relative paths. The actions and reusable workflows are added as well. This will allow one to reuse the workflows in all the branches.


surite (changes requested):
Second commit check failed

surite (approved):

PR #2274, opened 2023-04-18
fix bug 2255: [BUG] [JAVA Client] some i18n API URL fails with java.lang.NullPointerException

huihuiw01:


Xiaochao8 (dismissed):

tigershi (dismissed):

linr211:
I am wondering how this issue can be produced from client side if we have a specification for the URL?

tigershi (approved):

huihuiw01:
> I am wondering how this issue can be produced from client side if we have a specification for the URL?

This issue produced when customer do security scan on loginsight product which will send some unexpected URL.

These codes have been integrated into client loginsight use, but didn't merge to open source repo at that time. Currently loginsight will switch to open source version, so need merge these codes.

linr211 (approved):

PR #2334, opened 2023-05-22
add the country flag function

tigershi:


jin-jasper (approved):

linr211:
Is it better to include the repo https://gitlab.com/catamphetamine/country-flag-icons.git as gradle dependency so that it doesn't need to submit the svg files into Singleton repo which is hard to maintain the updates.

tigershi:
> Is it better to include the repo https://gitlab.com/catamphetamine/country-flag-icons.git as gradle dependency so that it doesn't need to submit the svg files into Singleton repo which is hard to maintain the updates.

I have confirmed with Jasper, he said use the current data can meet requirement. I also will investigate your comments

jin-jasper (approved):

g11n-ws/modules/md-service-i18n-l2/src/main/java/com/vmware/i18n/l2/dao/image/CountryFlagDaoImpl.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Map;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+@Repository
+public class CountryFlagDaoImpl implements ICountryFlagDao {
+    private static Logger logger = LoggerFactory.getLogger(CountryFlagDaoImpl.class);
+    private static String basePath = ConstantsKeys.IMAGE + File.separator + ConstantsKeys.FLAGS + File.separator;
+
+    @PostConstruct
```
huihuiw01, line 37: This convertion executed at runtime, right? I have same comment as @linr211, please move these code to g11n-cldr-pattern branch as other l2 patterns be converted before pattern build publish.

g11n-ws/modules/md-restful-api-i18n/src/main/java/com/vmware/vip/i18n/api/v2/image/ImageAPI.java
```diff
@@ -0,0 +1,61 @@
...
+    @Autowired
+    private ICountryFlagService countryFlagService;
+    /**
+     * Get country flag with region and level parameter.
+     */
+    @Operation(summary = APIOperation.IMAGE_COUNTRY_FLAG_VALUE, description = APIOperation.IMAGE_COUNTRY_FLAG_NOTES)
+    @GetMapping(value = APIV2.IMAGE_COUNTRY_FLAG_GET)
+    @ResponseStatus(HttpStatus.OK)
+    public void getCountryFlagWithRegion(
+            @Parameter(name = APIParamName.REGION, required = true, description = APIParamValue.REGION) @RequestParam(value = APIParamName.REGION, required = true) String region,
```
huihuiw01, line 43: 1. Must the 'region' parameter required? If the user need many countries' flags, he have to write many regions or request many times if only support one region in one request. So suggest it optional, and return all regions' flag if empty.

2. Can the 'region' parameter include multiple countries? If not, suggest it can allow multiple countries in case the caller would like to get multiple countries' images by one request considering the UI performance.

3. No validation for it?

g11n-ws/modules/md-service-i18n-l2/src/main/java/com/vmware/i18n/l2/dao/image/CountryFlagDaoImpl.java
tigershi: It doen't cldr data, so that we can't move data to cldr branch

g11n-ws/modules/md-restful-api-i18n/src/main/java/com/vmware/vip/i18n/api/v2/image/ImageAPI.java
tigershi: 1. Yes, we need the region parameter,  No requirement need to all flags. Could you have any use case about it.
2. No region include the multiple counties
3. We have validation in validation interceptor

huihuiw01 (approved):

PR #2347, opened 2023-05-29
Resolve security issues from Security scan

tigershi:


Xiaochao8 (approved):

linr211:
Seems the scope of the code changes are far beyond the security related, e.g. some name changes(like changing 'scopeFilter' to 'scopeFilterStr') are not security related. If so, could you separate the changes as security related and not related, or setup a call to go over the change? thanks!

tigershi:
we will create an other task to update the validate flow and filter scope flow

PR #2411, opened 2023-06-14
Fixed the Overly permissive regular expression range issue

tigershi:


linr211 (approved):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/utils/RegExpValidatorUtils.java
```diff
@@ -129,6 +129,13 @@ public static boolean containsHTML(String inputStr) {
 		return false;
 	}
 
+	public static boolean startLetterAndCommValidchar(String str) {
```
huihuiw01, line 132: Is it to validate if string start with letter and comma? 
If yes, suggest change the method name to 'startWithLetterAndComma' is more readable.

tigershi: No, only start letter, the other method is the same style

huihuiw01 (approved):

PR #2478, opened 2023-07-12
Updating Java version from 8 to 17, Spring Boot 3.1 has a minimum req…

Aleksbogdanov12:
Updating java version from 8 to 17, because Spring Boot 3.1 has the minimum requirement of Java 17. Calling reusable actions from devops branch, pre-process, check-header. Updating all places where there is code checkout with v3. Running git clone with depth 1 to increase speed of the workflow. Adding cronjob to run workflow once every month. The idea is that there are often new updates that are introduced in CodeQL, so it is a good idea, even if there are no new releases and pushes to the code, that it isscanned regularly.


surite (approved):

tigershi:
it doesn't need to upgrade the JDK, just fixed the

PR #2572, opened 2023-08-21
add the request body size

tigershi:


huihuiw01 (approved):

g11n-ws/vip-manager-l10n-agent/src/main/java/com/vmware/l10agent/conf/PropertyConfigs.java
```diff
@@ -197,4 +200,8 @@ public String getSyncListPath() {
 	public int getSyncBatchSize() { return (syncBatchSize - 1); }
 
 	public boolean isSyncBatchEnable() { return syncBatchEnable; }
+
+	public int getSyncReqBodySize(){
+		return  (1073741824*Integer.valueOf(this.reqBodySizeStr.toUpperCase().replaceAll("M", "").trim()))-536870912;
```
huihuiw01: Looks like the multiple 1073741824 is not right, should be 1048576(1024*1024), and suggest write it as 1024*1024, which is more readable. And same comment with 536870912.

huihuiw01 (approved):

g11n-ws/vip-manager-l10n-agent/src/main/java/com/vmware/l10agent/conf/PropertyConfigs.java
tigershi: done

g11n-ws/vip-manager-l10n-agent/src/main/java/com/vmware/l10agent/conf/PropertyConfigs.java
```diff
@@ -197,4 +200,8 @@ public String getSyncListPath() {
 	public int getSyncBatchSize() { return (syncBatchSize - 1); }
 
 	public boolean isSyncBatchEnable() { return syncBatchEnable; }
+
+	public int getSyncReqBodySize(){
+		return (1024*1024*Integer.valueOf(this.reqBodySizeStr.toUpperCase().replaceAll("M", "").trim())) - (512*1024);
```
linr211, line 205: it's better put the result of calculation to a variable so that it won't perform the calculation for each function call.

linr211 (approved):

g11n-ws/vip-manager-l10n-agent/src/main/java/com/vmware/l10agent/conf/PropertyConfigs.java
Xiaochao8, line 205: Why minus (512*1024)？

tigershi: OK

PR #2583, opened 2023-08-22
create service information file

Xiaochao8:


assets/bindata_generator.go
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright 2023 VMware, Inc.
```
tigershi, line 2: the copyright stable format is start with "Copyright 2022-2023 VMware, Inc."

Xiaochao8: For a new created file, this is the correct format. The format you provided is for the files which were created before current year.

tigershi (approved):

jin-jasper (approved):

PR #2598, opened 2023-08-25
add the Versionlist API

tigershi:


g11n-ws/modules/md-restful-api/src/main/java/com/vmware/vip/api/rest/APIV2.java
```diff
@@ -18,6 +18,7 @@ public class APIV2 {
 	public static final String PRODUCT_VERSIONINFO_GET       = API_TRANSLATOIN + "/products/{"+APIParamName.PRODUCT_NAME+"}/versions/{"+APIParamName.VERSION2+"}/versioninfo";
 	public static final String PRODUCT_LOCALE_LIST_GET    = API_TRANSLATOIN + "/products/{"+APIParamName.PRODUCT_NAME+"}/versions/{"+APIParamName.VERSION2+"}/localelist";
 	public static final String PRODUCT_COMPONENT_LIST_GET = API_TRANSLATOIN + "/products/{"+APIParamName.PRODUCT_NAME+"}/versions/{"+APIParamName.VERSION2+"}/componentlist";
+	public static final String PRODUCT_VERSION_LIST_GET   = API_TRANSLATOIN + "/products/{"+APIParamName.PRODUCT_NAME+"}/versionlist";
```
linr211, line 21: Instead of creating this new API, why not change existing version API(`/i18n/api/v2/about/version`) to fit this case? e.g. we can have the version parameter as optional, if it's not provided then respond with the version list, otherwise respond with the specific version's details.

tigershi: this is different requirement，I will check your suggest. thanks

Xiaochao8 (approved):

g11n-ws/modules/md-restful-api/src/main/java/com/vmware/vip/api/rest/APIV2.java
tigershi: I have checked the about version API, The main purpose of about version API is for product team to track Singleton service upgrade status and translation version. The version list API is get  all available version list.  if we put the available list in about version API, it will be confusion. So that, I set the API name is versionlist

linr211 (approved):

PR #2696, opened 2023-09-22
support to get translation of the same key in multiple versions

Xiaochao8:


tigershi:
unit test failure

Xiaochao8:
> unit test failure

This is known as a CI issue.

PR #2699, opened 2023-09-25
 Optimize response code/message for l10n response

tigershi:


Xiaochao8 (dismissed):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/i18n/status/APIResponseStatus.java
```diff
@@ -33,4 +33,6 @@ public class APIResponseStatus {
             "Not Found");
     public static final Response UNKNOWN_ERROR = new Response(520, "Unknown Error");
     public static final Response TRANSLATION_COLLECT_REQUEST_SUCCESS = new Response(200, "Source collection request have received");
+    public static final Response TRANSLATION_COLLECT_SOURCE_PARAM_IS_FAlSE = new Response(200, "Received the sources and comments(please use translation-product-component-api to confirm it).");
```
huihuiw01, line 36: Is this message correct? When 'collectSource=false', still can receive the sources?

tigershi: No, It just let client use the  translation-product-component-api to confirm it

huihuiw01 (dismissed):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/i18n/status/APIResponseStatus.java
```diff
@@ -32,5 +32,7 @@ public class APIResponseStatus {
     public static final Response INTERNAL_NO_RESOURCE_ERROR = new Response(404,
             "Not Found");
     public static final Response UNKNOWN_ERROR = new Response(520, "Unknown Error");
-    public static final Response TRANSLATION_COLLECT_REQUEST_SUCCESS = new Response(200, "Source collection request have received");
+    public static final Response TRANSLATION_COLLECT_REQUEST_SUCCESS = new Response(200, "Source collection request have received(please use translation-product-component-api to confirm it)");
+    public static final Response TRANSLATION_COLLECT_SOURCE_PARAM_IS_FAlSE = new Response(200, "Check received the sources status, please use translation-product-component-api to confirm it.");
```
Xiaochao8, line 36: Should the response code be 400 and the message be "the 'collectSource' is false and won't do source collection"?

Xiaochao8 (approved):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/i18n/status/APIResponseStatus.java
```diff
@@ -32,5 +32,7 @@ public class APIResponseStatus {
     public static final Response INTERNAL_NO_RESOURCE_ERROR = new Response(404,
             "Not Found");
     public static final Response UNKNOWN_ERROR = new Response(520, "Unknown Error");
-    public static final Response TRANSLATION_COLLECT_REQUEST_SUCCESS = new Response(200, "Source collection request have received");
+    public static final Response TRANSLATION_COLLECT_REQUEST_SUCCESS = new Response(200, "Source collection request have received(please use translation-product-component-api to confirm it)");
+    public static final Response TRANSLATION_COLLECT_SOURCE_PARAM_IS_FAlSE = new Response(400, "Check received the sources status, please use translation-product-component-api to confirm it.");
```
huihuiw01, line 36: If change the code to 400, that means 'false' is invalid value, only support value 'true', and from the bug's description this is required. 

So I think change the response code to '400' and message to 'Incorrect collectsource(only allows true)' is enough. So you need revert the change for message COLLECTSOURCE_NOT_VALIDE_L10N and other changes accordingly.

Besides you can set it as 'true' by default.

huihuiw01 (changes requested):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/i18n/status/APIResponseStatus.java
tigershi: I change the code to 200, The COLLECTSOURCE_NOT_VALIDE can't change

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/i18n/status/APIResponseStatus.java
```diff
@@ -32,5 +32,7 @@ public class APIResponseStatus {
     public static final Response INTERNAL_NO_RESOURCE_ERROR = new Response(404,
             "Not Found");
     public static final Response UNKNOWN_ERROR = new Response(520, "Unknown Error");
-    public static final Response TRANSLATION_COLLECT_REQUEST_SUCCESS = new Response(200, "Source collection request have received");
+    public static final Response TRANSLATION_COLLECT_REQUEST_SUCCESS = new Response(200, "Source collection request have received(please use translation-product-component-api to confirm it)");
+    public static final Response TRANSLATION_COLLECT_SOURCE_PARAM_IS_FAlSE = new Response(605, "Check received the sources status, please use translation-product-component-api to confirm it.");
```
huihuiw01, line 36: What does 605 mean?

Besides I think you should point out "doesn't do source collection since 'collectSource' is false", contrary to TRANSLATION_COLLECT_REQUEST_SUCCESS.

huihuiw01 (changes requested):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/i18n/status/APIResponseStatus.java
tigershi: Different format response

huihuiw01 (dismissed):

g11n-ws/vip-common/src/main/java/com/vmware/vip/common/constants/ValidationMsg.java
```diff
@@ -22,7 +22,7 @@ public final class ValidationMsg {
 	public final static String COMBINE_NOT_VALIDE = "Incorrect combine(only 1 or 2)";
 	public final static String SCOPE_NOT_VALIDE = "Incorrect scope(allows letter, comma)";
 	public final static String COLLECTSOURCE_NOT_VALIDE = "Incorrect collectsource(only allows true, false)";
-	public final static String COLLECTSOURCE_NOT_VALIDE_L10N = "Incorrect collectsource(only allows true)";
+	public final static String COLLECTSOURCE_NOT_VALIDE_L10N = "Incorrect collectSource(only allows true, please use translation-product-component-api to confirm received source status)";
```
Xiaochao8, line 25: Does it need to change this message?
The source isn't collected successfully, so don't need to tell users how to verify.

tigershi: Yes, because the requirement

linr211 (approved):

huihuiw01 (approved):

PR #2749, opened 2023-10-12
fix L10n source collection lost key

tigershi:


Xiaochao8 (commented):
What's the reason of lock failure?

PR #2955, opened 2024-01-08
add load remote git configuration feature

tigershi:


linr211:
is this feature related with the service itself? if no, i would suggest move it out of g11n-ws.

tigershi:
> is this feature related with the service itself? if no, i would suggest move it out of g11n-ws.

All spring boot APP that Compile compatible with JDK17

jin-jasper (approved):

Xiaochao8 (approved):

